PR #183, opened 2020-09-22
[sunsch] Prove bug 1948 in OPM, UDQ and lineshift

berland:
When https://github.com/OPM/opm-common/issues/1948 is fixed, the tests in this might pass.


ertomatic:

wouterjdb (approved):

Issue #244, opened 2020-11-23
[sunsch] Exported VFP tables may become too wide for Eclipse processing


asnyv:
The issue is probably in the dependency `opm-common` and not `sunsch` itself, but not sure if `OPM Flow` will struggle with it in the same way as Eclipse, and whether solving it is something the OPM project in that case would prioritize. Also, the OPM release schedule is not very frequent, and a fix cannot be expected in the next ~6 months.

Issue is this:
The lines written to file for VFP tables may become longer than what Eclipse seems to read, I haven't throughly investigated the limit yet, but it seems to be roughly 100 characters. This results in errors of missing / to finalize VFP input (and Eclipse not reading the full table). The export appears to add some line splitting, but my feeling is that it is not accounting for the number of digits in each value of the table, and instead splitting after n tables values, which may not work if the VFP table has many digits per entry. One may of course argue that the VFP tables aren't that precise anyways, and that the number of significant digits in the table could be reduced, but I think it should not be a limitation in sunsch / opm.

As of now I have bypassed it myself by postprocessing sunsch exports, searching for VFPPROD / VFPINJ and split lines when needed.

To do this in sunsch may of course be a bit "dirty" and ideally the `opm` `TimeVector` should return a string with appropriate line lengths.

PR #245, opened 2020-11-24
Solve bug in opm-common on too long lines.

berland:
Bandaid solution for sunsch before upstream fix is in place

Solves #244

### DIFF
--- a/src/subscript/sunsch/sunsch.py
+++ b/src/subscript/sunsch/sunsch.py
@@ -10,6 +10,7 @@ import os
 import datetime
 import tempfile
 import argparse
+import textwrap
 import logging
 
 import yaml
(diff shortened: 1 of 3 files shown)


src/subscript/sunsch/sunsch.py
```diff
@@ -551,6 +552,48 @@ def substitute(insert_statement):
...
+        warn (bool): Whether to log a warning for each line
+            of excessive length
+
+    Returns:
+        str: Multiline string with no lines more than maxchars
+        in length. Trailing whitespace is always stripped.
+    """
+    wrappedstr = ""
+    for line_idx, line in enumerate(string.split("\n")):
+        commentsplit = line.split("--")
```
asnyv, line 572: Some may have more than one set of `--` (if they are eager on commenting in and out), suggest to use partition instead of split, or add the comment again further down as `"--".join(commentsplit[1:]"`

asnyv (changes requested):
One comment, otherwise LGTM 👍

src/subscript/sunsch/sunsch.py
asnyv: Or `line.split("--", 1)` would also do the trick

berland: `partition` made it prettier.

asnyv (approved):
LGTM 😊

Issue #247, opened 2020-11-25
PRTVOL2CSV fails when mix of names and numbers defined in region2fipnum.


mareco701:
when having a mix of names and numbers in the region2fipnum.yaml file (see example below where one of the region is named by a number) then the script fails with following error message which is not really explicit:
 
![image](https://user-images.githubusercontent.com/70753366/100238723-d0073100-2f30-11eb-9544-141348f6dc87.png)

![image](https://user-images.githubusercontent.com/70753366/100238795-e9a87880-2f30-11eb-9765-0212637f56b5.png)

should be fixed or just get a clearer error message

berland:
@mareco701 , are  you able to  test if this bug  has been resolved? (through virtual env or komodo bleeding)

mareco701:
Hi, yes, I tested it and it now generates a file named by default rft_ertobs_sim.csv.

mareco701:
Hi, sorry the answer before was not for this... but I have now tested prtvol2csv with the regions mix set which failed but now works fine with the fix.

PR #248, opened 2020-11-26
Ensure regions in prtvol2csv are always strings

berland:
Potentially solves #247, even though the added test did not trigger error in development system. The issue might relate to specific pandas vs yaml versions.


wouterjdb (approved):
LGTM

PR #253, opened 2020-12-02
Fix default for MERGE_RFT_ERTOBS forward model

berland:
Avoid writing to a filename <OUTPUT> by default.

Add OUTPUT to documentation example for forward model

Also set default output filename for command line script.

Credit: @mareco701


rnyb (approved):

mareco701:
hi, I guess the default name should be the one webviz is looking for i.e. rft_ert.csv.

mareco701 (commented):
Consider changing the default output filename to share/results/tables/rft_ert.csv so this is in line with default input expected in webviz (name and location) ? if only the csv file name is entered then the file is not created at the correct place (under the correct folder).

berland:
Thanks. This has been considered, but there is another recommendation in the FMU-sphere that tools should not create directories implicitly. Given that, it is a little uncomfortable to default to a file in a directory that must be assumed to exist upfront.

berland:
The default name of the output file is a little bit more tricky and of course up for discussion. The default here is according to the name of the tool itself (and also a new tool in fmu-tools named after this particular one). I find  `rft_ert.csv` a little less descriptive than `rft_ertobs.csv`, but neither filenames are particularly good.

mareco701 (approved):
I see,  thanks for the explanation.

mrec:
@berland - I don't know who you meant to credit in the top comment, but it definitely wasn't me!

berland:
> @berland - I don't know who you meant to credit in the top comment, but it definitely wasn't me!

Sorry, your user name on github matches an internal  username in Equinor..

PR #255, opened 2020-12-03
Port from travis to github actions, incl gh-pages

berland:


eivindsm (approved):
This is giberish, but looks good to me. Maybe akia, can give a better review if need.

berland:
@wouterjdb has been summoned, but I trust your instincts on giberish.

Issue #192, opened 2020-09-28
ERT forward model hook for csv_stack


berland:
An ERT forward model hook for csv_stack should be made.

berland:
And a workflow job.

PR #256, opened 2020-12-07
[csv_stack] Refactor, ERT integration

berland:
Solves #192


ertomatic:

asnyv:
Is this a review request, or am I actually assigned to do something @berland? 😅

berland:
Hereby assigned to do some reviewing ;)

src/subscript/csv_stack/csv_stack.py
```diff
@@ -17,42 +98,43 @@ class CustomFormatter(
...
+        help="Input CSV file. If you use -, it will read from stdin ",
     )
     parser.add_argument(
         "-o",
         "--output",
         type=str,
         help=(
-            "name of output csv file. "
-            "Use - or stdout to have the output dumped to stdout."
+            "Name of output csv file. " "Use - or to have the output dumped to stdout."
```
wouterjdb, line 130: Two strings with a space?

wouterjdb (commented):
- Some functions have type annotation, some have not.

I'm not really familiar enough with this script to "quickly" do a review on its functionality though.

src/subscript/csv_stack/csv_stack.py
berland: Fixed

src/subscript/csv_stack/csv_stack.py
```diff
@@ -1,7 +1,88 @@
...
+"""  # noqa
+
+CATEGORY = "utility.transformation"
+
+# The following string is used for the ERT workflow documentation, note
+# the very subtle difference in variable name.
+WORKFLOW_EXAMPLE = """
+Add a file named e.g. ``ert/bin/workflows/CSV_STACK_WELLS`` with the contents::
+
+  CSV_STACK \"<CASEDIR>/share/results/tables/unsmry--montly.csv\" "--split" well "--keepminimal"
```
asnyv, line 47: You are more familiar with ERT workflow syntax than me, but are the `\`'s necessary?

Also, should we allow `-s` as alias for `--split` and `-k` for `--keepminimal` in the parser to avoid the need of `"`?
Or do you prefer to be explicit? I'm not sure myself...

src/subscript/csv_stack/csv_stack.py
```diff
@@ -65,133 +145,198 @@ def get_parser():
     parser.add_argument(
         "--keepminimal",
         action="store_true",
-        help="Keep only REAL, ENSEMBLE, DATE and unpivoted columns",
+        help=(
+            "Keep only REAL, ENSEMBLE, DATE and unpivoted columns. "
+            "Implies dropping constant columns"
+        ),
```
asnyv, line 151: this list doesn't ceover the whole `ALWAYS_KEEP` list. Of course, these are the most likely ones from export with e.g. `CSV_EXPORT` / f`mu-ensemble`, but a bit misleading

src/subscript/csv_stack/csv_stack.py
berland: I think you are right, the backslashes are probably a remnant. It is not for ERT, but for Python.  Will remove

src/subscript/csv_stack/csv_stack.py
```diff
@@ -1,7 +1,88 @@
...
+#  0: regexp for matching in front of the colum name separator
+#  1: Separator in column names
+#  2: Name of new column
+STACK_LIBRARY = {
+    "well": ["W[A-Z0-9]*:.*", ":", "WELL"],
+    "region": ["R[A-Z_0-9]*:.*", ":", "REGION"],
+    "group": ["G[A-Z0-9]*:.*", ":", "GROUP"],
+    "block": ["B[A-Z0-9]*:.*", ":", "BLOCK"],
+    "all": [".*:.*", ":", "IDENTIFIER"],
+}
```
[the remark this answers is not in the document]
berland: Yes, add to new issue if wanted.

src/subscript/csv_stack/csv_stack.py
berland: Added `-s`, but `-k` would be ambiguous. Keeping long options in the example for the sake of being explicit and as a user remainder to quote double dashes.

src/subscript/csv_stack/csv_stack.py
berland: Fair enough. Removing the legacy column names from `ALWAYS_KEEP`. Implies an API break, but anyone relying on this would be a big surprise.

asnyv (approved):
LGTM 👍

Issue #258, opened 2020-12-07
[fmuobs] Support MM.DD.YYYY, not only MM/DD/YYYY


eivindsm:
Parsing of dates in ertobs files are done incorrect. ERT syntax is DD.MM.YYYY, but this is parsed as MM.DD.YY. As illustrated below. Output from fmuobs, follwed by the input. 

(kenv_2020.11 + 2020.11.00-py36-rhel7) eism@st-linrgsn198:/project/fmu/users/eism/reek_hm/r006/ert/input/observations> more wwct.yml 
smry:
- key: WWCT:OP_4
  observations:
  - date: '2001-01-12'
    error: 0.1
    label: OP_4_WWCT1
    value: 0.17
  - date: '2002-01-10'
    error: 0.1
    label: OP_4_WWCT2
    value: 0.43
  - date: '2002-01-12'
    error: 0.1
    label: OP_4_WWCT3
    value: 0.47
(kenv_2020.11 + 2020.11.00-py36-rhel7) eism@st-linrgsn198:/project/fmu/users/eism/reek_hm/r006/ert/input/observations> more wwct_op5.txt 
-- WWCT data

SUMMARY_OBSERVATION OP_4_WWCT1 {VALUE=0.17; ERROR=0.1; DATE=01.12.2001; KEY= WWCT:OP_4;};
SUMMARY_OBSERVATION OP_4_WWCT2 {VALUE=0.43; ERROR=0.1; DATE=01.10.2002; KEY= WWCT:OP_4;};
SUMMARY_OBSERVATION OP_4_WWCT3 {VALUE=0.47; ERROR=0.1; DATE=01.12.2002; KEY= WWCT:OP_4;};

berland:
The ERT doc on observations only mentions DD/MM/YYYY,  not DD.MM.YYYY as date format, so that MM.DD.YYYY does not  error hard is a bonus (or bug in itself).  Adding MM.DD.YYYY is a feature request.

PR #260, opened 2020-12-07
Support alternative date format in fmuobs

berland:
Solves #258


eivindsm (approved):
Perfect! Thanks for supporting the undocumented date format in ERT 👍

PR #265, opened 2020-12-09
Add OPM as a direct dependency

berland:
At least `sunsch` uses OPM directly (other tools mostly through ecl2df)


markusdregi (approved):
👍

PR #268, opened 2020-12-11
Add testing via komodo

eivindjahren:
Adds running tests in komodo bleeding on pull requests. This can be extended in the future to running the tests against an installed version of new releases of komodo in the future.


berland (approved):
LGTM

PR #271, opened 2020-12-15
Explain diff_dates.txt syntax better in ecldiff2roff

berland:
Solves #267


rnyb (changes requested):
Looks good, but could consider adding to description that the diff dates are space seperated

berland:
> Looks good, but could consider adding to description that the diff dates are space seperated

will fix

rnyb (approved):

PR #272, opened 2020-12-16
Support report_step in merge_rft_ertobs

berland:
Extracts in a non-robust way the report_step from the observation filename.
The report_step is not important for the CSV file this script produces, but getting
the well-name right is. create_rft_ertobs.py from fmu-tools creates obs filenames
on this syntax.

Also fix the index of a returned dataframe, invisible in the output from
this script.


src/subscript/merge_rft_ertobs/merge_rft_ertobs.py
```diff
@@ -59,13 +67,58 @@ def get_parser():
...
+        A-5_99 gives (A-5_99, 1)  # report steps more than 10 not supported.
+        "R_A4_1" gives (R_A4, 1)
+        "R_A4" gives (R_A4, 1)
+        "R_A_4" gives (R_A, 4)  # Warning, this is probaly unintended!
+
+    Args
+        wellname_reportstep (str)
+
+    Returns
+        tuple: wellname and reportstep. Reportstep defaulted to 1 if not found
```
wouterjdb, line 96: Why 1 and not None?

src/subscript/merge_rft_ertobs/merge_rft_ertobs.py
```diff
@@ -101,9 +154,10 @@ def get_observations(obsdir="", filepattern="*.obs"):
     for obsfilename in glob.glob(os.path.join(obsdir, filepattern)):
         # Warning: Deducing the wellname this way will fail for
         # exotic filepatterns.
-        wellname = os.path.basename(obsfilename).split(filepattern.replace("*", ""))[
-            0:-1
-        ][0]
+        wellname_reportstep = os.path.basename(obsfilename).split(
+            filepattern.replace("*", "")
```
wouterjdb, line 158: Isn't the `*` already gone due to `obsfilename` being a result from `glob.glob`?

src/subscript/merge_rft_ertobs/merge_rft_ertobs.py
berland: It must/should match the default number in use in the ert config, which is never None.

src/subscript/merge_rft_ertobs/merge_rft_ertobs.py
berland: It is gone in `obsfilename`, but here we remove it from the filepattern in order to chop away the extension. I see `split()` could have been replace by `replace` also, but the way the code works here is suboptimal in any case (relying on filenames).

asnyv (approved):
LGTM 👍

rnyb (commented):
As you've already mentioned, this may be a pitfall:
"R_A_4" gives (R_A, 4)  # Warning, this is probaly unintended!

Would it make sense to add a check if there's 1 or more numbers and if only one number it returns reportstep1?

berland:
> As you've already mentioned, this may be a pitfall:
> "R_A_4" gives (R_A, 4) # Warning, this is probaly unintended!
> 
> Would it make sense to add a check if there's 1 or more numbers and if only one number it returns reportstep1?

Maybe, but I am not sure if that would only be a marginal improvement, and that it would not create other issues from trying to be too smart (and new  bugs would become more difficult to understand), so this is a difficult trade-off ;)

The underlying problem is that extracting metadata for each observation from a filename is fundamentally flawed, but not likely to be fixable before earliest ERT3.

PR #274, opened 2020-12-16
Refactor prtvol2csv to use a new FipMapper class

berland:
The FipMapper class is a first version of an API that may
later move out of subscript (fmu-config is a candidate).

Also deprecate the region export, and the implicit mkdir.
Both these feature should be taken out at some point

### DIFF
--- a/src/subscript/prtvol2csv/fipmapper.py
+++ b/src/subscript/prtvol2csv/fipmapper.py
@@ -1,0 +1,245 @@
+"""The FipMapper class, mapping region/zones in RMS to FIPxxx in Eclipse.
+
+This API should be considered private to prtvol2csv until it has moved somewhere
+else"""
+import yaml
+from typing import Union
+from pathlib import Path
+
+from subscript import getLogger
+
+logger = getLogger(__name__)
+
+
+class FipMapper:
+    def __init__(
+        self,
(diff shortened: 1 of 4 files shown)


src/subscript/prtvol2csv/fipmapper.py
```diff
@@ -0,0 +1,235 @@
...
+        Several data structures in the dictionary can be used, such that
+        the needed information can be extracted from the global configurations
+        file.
+
+        Args:
+            yamlfile (Path): Filename
+            mapdata (dict): direct dictionary input. Provide only one of the
+                arguments, not both.
+            skipstring: List of strings which will be ignored (e.g. "Totals").
+        """
```
jcrivenaes, line 32: (e.g. ["Totals"])

jcrivenaes (approved):
OK for me

rnyb (commented):
Not sure i see the benefit of removing the implicit mkdir.

berland:
> Not sure i see the benefit of removing the implicit mkdir.

IT has concluded that it  is best practice not to be implicit about mkdir's.

rnyb:
> > Not sure i see the benefit of removing the implicit mkdir.
> 
> IT has concluded that it is best practice not to be implicit about mkdir's.

I see -  i don't understand the rationale behind it though

PR #283, opened 2021-01-08
Fix bug with testpylib in runrms

berland:
Underscore-inconsistencty between python code and deployed yml-file, and
added support for both list and string input syntax in yaml field pythonpathtest


ertomatic:

jcrivenaes (approved):
Looks good

Issue #196, opened 2020-09-29
Route INFO  and WARNING to stdout, ERROR to stderr


berland:
The subscript logger routes all logging to stderr, which confuse users and applications (ERT) that reads both stdout and stderr files.

There is no timestamp in the current logger. That means that if the logger messages are split into two files, patching them together will be difficult  since the order of messages will not be obvious.  This is a problem for example a  problem if some scripts emit ERROR messages but do not quit immediately

berland:
> A solution that fixes your "difficult to patch" would be to route `INFO`, `WARNING` and `ERROR` to stdout, but only `ERROR` to `stderr`.

When run interactively in a terminal, I suppose this will imply `ERROR` messages to appear twice?

PR #284, opened 2021-01-11
Let ertwatch only display the tail of stderr files

berland:
Minor fix for the terminal users out there.

`tail` instead of `cat` of stderr files is to some degree caused by subscript logging writing INFO to stderr, see issue #196


ertomatic:

eivindsm (approved):
LGTM

PR #290, opened 2021-01-14
Test sw model utilities

berland:


ertomatic:

berland:
Commit 41d0f78 was included here by an error, but should be fine as long as this PR is merged and not squashed.

jcrivenaes (approved):

PR #294, opened 2021-01-20
Bugfix prtvol2csv for non-existing fipnums

berland:
The deprecated region export crashes if yaml files
mention fipnums that are empty/missing


ertomatic:

rnyb (approved):

PR #298, opened 2021-01-25
Fix tests on komodo deploy

eivindjahren:
In order to test installed packages in komodo, we need to install test requirements without installing the package. For that we need to split test requirements into requirement.txt files.

### DIFF
--- a/docs_requirements.txt
+++ b/docs_requirements.txt
@@ -1,0 +1,5 @@
+
+sphinx
+sphinx-argparse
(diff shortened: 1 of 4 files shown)


docs_requirements.txt
```diff
@@ -0,0 +1,5 @@
+
```
berland, line 1: empty line?

berland (approved):
ok!

PR #303, opened 2021-02-04
Enforce deprecation for sunsch v1 config files

berland:
With this PR, sunsch will fail hard when given a V1 config file, but it will still print out an auto-converted V2 config. 

V1 config file format has been deprecated since June 2020.

Suggest to wait some more months before pruning all v1 support code.


eivindsm (approved):
Supported

PR #313, opened 2021-02-17
Change column name WATER_PORV to WATPV_TOTAL in prtvol2csv

berland:


berland:
@asnyv , is Webviz using this CSV output directly and does it need a change as well?

asnyv:
> @asnyv , is Webviz using this CSV output directly and does it need a change as well?

Nope, at least not as far as I know. Some of the other column names are used to add some descriptive text (but they are mainly based on the RMS Volumetrics job output). Will at least be very suprised if this change is breaking anything in webviz-subsurface.

Any other motivation behind this change than to keep the same format as for oil and gas? 
Asking because I'm not sure myself: Are all of these names consistent with the corresponding column names (if they exist) that the RMS 3D Volumetrics job exports? Nice to keep the same convention to make it easy to compare RMS output with initialized Eclipse volumes (though the RMS calculations are often quite simplified).

rnyb (approved):

rnyb:
We should also consider renaming WATER_TOTAL to WIIP_TOTAL  (WIIP = Water Initially In Placce) 

(equivalent to  the  oil STOIIP_TOTAL and gas  GIIP_TOTAL columns)

berland:
https://github.com/equinor/ecl2df/pull/271 changes WATER_TOTAL to WIIP_TOTAL.

PR #314, opened 2021-02-19
Support having parameters.txt in the same dir

berland:
Better fit with ert3

Added test functions


eivindsm (commented):
Places to check for files is set to [.,../..], should we also check for files in .. ?

eivindsm (approved):
Perfect :)

PR #317, opened 2021-02-26
Pathlib and pytest-mocker all over

berland:


eivindsm (approved):
Wow, a big commit! LGTM

PR #333, opened 2021-03-19
Split log output to stdout and stderr

berland:
* Also fix log level bug in merge_schedule that tests uncovered.

### DIFF
--- a/src/subscript/__init__.py
+++ b/src/subscript/__init__.py
@@ -1,3 +1,4 @@
+import sys
 import logging
 
 try:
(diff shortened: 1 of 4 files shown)


src/subscript/__init__.py
```diff
@@ -24,13 +38,43 @@ def getLogger(module_name="subscript"):
...
+    def filter(self, record):
+        return record.levelno >= logging.ERROR
+
+
+class StdoutFilter(logging.Filter):
+    """Used to filter out log records from WARNING and below"""
+
+    # pylint: disable=too-few-public-methods
+    def filter(self, record):
+        return record.levelno < logging.ERROR
```
jondequinor, line 80: You could replace these with lambdas.

`stderr_handler.addFilter(lambda record: record.levelno >= logging.ERROR)`

jondequinor (approved):
LGTM! But be warned I spent only a few minutes reading about this.

PR #335, opened 2021-03-24
Adapt logger to be reused in subscript_internal

berland:
Subscript_internal  should be able to use the subscript logger identical to how subscript scripts use it.


tklov11 (approved):
LGTM

PR #340, opened 2021-04-19
Support gas-water decks in check_swatinit

berland:
This ensures rudimentary support for GAS - WATER DATA decks (solves test code failure)

Requires Eclipse100 to test.


vkip (approved):
lgtm :)

Issue #342, opened 2021-04-21
[check_swatinit] smalll things


berland:
* Errors if EQLNUM > 1 in the grid, conflict with the default probably. ERror message occurs at the end

* is/are typo in caption in docs

* waterfall plot: when numbers are negative, the blue text is not visible on blue background

* last column in waterfall plot is not correct with large negatives.

* printing large tables to sdtout gives dots, pandas should not compress the output. fix

PR #343, opened 2021-04-22
Fix small issues with check_swatinit after user testing

berland:
Solves #342 

* Fixed palette for scatter plots
* Regenerated and almost reproducible plots in doc
* Adjust placement of annotations in volplot
* Reorder arguments in parser, for --help output
* Allow EQLNUM=1 to not exist without crash
* Ensure large dataframes are printed in full to stdout.
* Add interactive plots to pytest, not run by default
* Add SWU to output csv and notify that SWU<1 is not supported


vkip (approved):
LGTM

PR #346, opened 2021-04-26
Fix docstring whitespaces

berland:
Fixed and enforced by black 21.4b0


eivindsm (approved):
Looks good; simplest RP review ever!

PR #328, opened 2021-03-04
Adding ri_wellmod - ResInsight well modelling wrapper

vkip:

PR #350, opened 2021-04-28
Adding ri_wellmod - ResInsight well modelling wrapper

berland:
Adding drogon well trajectories

Adding UNRST (currently needed for LGR test)

Renaming to *_NOSIM and adding sim version

Adding LGR NOSIM case

Saving LGR test cases and tests

Added ERT job

Added RI_WELLMOD to expected job list in test_hook_implementations

Added ERT integration test

Adding check for ResInsight also in ERT integration test

Adding Reek tests

Reek LGR test only for on-site system

rename RUNPATH, pick up lower-cased-resinsight-binary


berland:
Supersedes #328

berland:
The argparser defines a mixture of short and long options. Can the hybrid option  names "-msw", "-lo", "-tmp" be dropped without breaking existing stuff?

The option "-tmp" is ambigous as it could also be equivalent to "-t -m -p", therefore one should stick to one-letter short options with one dash, and multi-letter long options with double dash.

vkip:
> b

Yes, suggest to have --msw as the long option name, this requires minor changes, for the rest the short options may simply be removed. On it now..

tests/test_ri_wellmod.py
```diff
@@ -257,3 +258,39 @@ def test_main_lgr_reek(tmpdir, mocker):
     ri_wellmod.main()
 
     assert Path(outfile).exists() and file_contains(outfile, "OP_1")
+
+
+# Test development version
+@pytest.mark.skipif(drogon_runpath() is None, reason="Could not find Drogon data")
```
berland, line 264: add
`@pytest.mark.skipif(not Path(RI_DEV).exists(), reason="No dev-version of ResInsight available")`

src/subscript/ri_wellmod/ri_wellmod.py
```diff
@@ -513,7 +521,32 @@ def main():
         logger.debug("Command line parameters are: %s", command_line_parameters)
 
     # Launch ResInsight
-    resinsight = launch_resinsight(console_mode, command_line_parameters)
+    if resinsightdev:  # Use development version
```
berland, line 524: Move to dedicated python function? 

`main()` is already too long..

tests/test_ri_wellmod.py
vkip: Good point, done.

eivindsm (approved):
Looks really good, no objections to this PR

PR #352, opened 2021-04-28
[fmuobs] Add naive support for GENERAL_OBSERVATIONs in yaml

berland:
Implements the most obvious/naive way of converting ert observations to the yaml format, meaning that

```
GENERAL_OBSERVATION GEN_OBS1{
   DATA     = SOME_FIELD;
   RESTART  = 20;
   OBS_FILE = some_file.txt;
};

GENERAL_OBSERVATION GEN_OBS2{
   DATA       = SOME_FIELD;
   INDEX_LIST = 0,3,9;
   RESTART    = 20;
   OBS_FILE   = some_file.txt;
};
```
is translated (in both directions) to:

```
general:
 - label: GEN_OBS1
   data: SOME_FIELD
   restart: 20.0
   obs_file: some_file.txt
 - label: GEN_OBS2
   data: SOME_FIELD
   index_list: "0,3,9"
   restart: 20.0
   obs_file: some_file.txt

```

### DIFF
--- a/src/subscript/fmuobs/parsers.py
+++ b/src/subscript/fmuobs/parsers.py
@@ -504,7 +504,7 @@ def smrydictlist2df(smrylist):
 
 def blockdictlist2df(blocklist):
     """Parse a list structure (subpart of yaml syntax) of block observations
-    into  dataframe format
+    into dataframe format
 
     The internal dataframe format uses "LABEL" and "OBS" as unique keys
     for individual observations. These are constructed if not present.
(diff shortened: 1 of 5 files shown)


eivindsm (commented):
I would expect the parser to read the obsfile, fetch the observation values and stds corrending to indexlist, and internalize these as two lists. Leaving behind a reference to a file does not serve much purpose. The format for these files is well defined, so it seems realistic to fetch the numbers.  Sorry, if I missed something obvious :)

berland:
Good point, parsing that text file is the original request. I am not yet sure if such parsing can be done in a reversible manner and it at least requires more thinking for how to convert back to ERT observation format. Should it create its own OBS_FILE on disk if it is converting from yaml back to ert-obs?

The current implementation is the obvious implementation by naivety. Parsing the OBS_FILE should be an extension of this.

eivindsm (commented):
I have no objections to the code, and can approve. However, I am unsure if I like the proposed yml- data structure; 
- The restart entry in ERT does really refer to restart numbering, but a required identifier that in may refer to restart (I think it did in good old EnKF days).
- Should the index list be a string and not a list? 
- Will we have to change the structure if add on parsing of numbers?
Another thought, is it feasible to output this into a format that can be used with eg WebViz or Resinsight?

berland:
Converted to draft PR. It is not important to merge this as it would be no use currently. We should think through what we want from this, and what is possible to achieve from a format converter.

PR #353, opened 2021-04-28
Deprecate merge_schedule

berland:
It  is a too thin wrapper around sunsch to deserve its own endpoint.


berland:
The user are now faced with this when she runs the script:14:49 $ merge_schedule -v merge2.sch schedule.sch  foo.sch
```
......merge_schedule/merge_schedule.py:54: FutureWarning: merge_schedule is deprecated, use sunsch
  warnings.warn("merge_schedule is deprecated, use sunsch", FutureWarning)
Add the following lines to a yaml file and use it as input for sunsch:
----
files:
- merge2.sch
- schedule.sch
output: foo.sch
----
```

src/subscript/merge_schedule/merge_schedule.py
```diff
@@ -50,14 +50,20 @@ def main():
...
+
+    sunsch_config = {"files": args.inputfiles}
 
     if args.verbose:
         # Set the root logger to INFO, will be inherited by sunsch
         logger.setLevel(logging.INFO)
 
-    logger.info("# Sending the following YAML configuration to sunsch:")
-    logger.info(yaml.dump(sunsch_config))
+    warnings.warn
```
anders-kiaer, line 61: What does this line do?

anders-kiaer (approved):

src/subscript/merge_schedule/merge_schedule.py
berland: That should not be there!

berland: lgtm.com, where were you..? ;)

PR #355, opened 2021-04-29
Solve bug on PPCWMAX grid merging

berland:
* Cell index was shuffled through pd.merge
* Add REGIONS output from PillarModel (missing), outputting SATNUM and EQLNUM.
* Adjust whitespace in PillarModel output


vkip (approved):
LGTM!

PR #362, opened 2021-05-04
Add type hints to most files, and add to CI

berland:


.github/workflows/subscript.yml
```diff
@@ -46,6 +46,11 @@ jobs:
           pip install flake8
           flake8 src tests
 
+      - name: Enforce static typing
+        run: |
+          pip install mypy
+          mypy src/subscript
```
jondequinor, line 52: Why not add this to test_requirements.txt or similar?

src/subscript/pack_sim/pack_sim.py
```diff
@@ -420,50 +434,72 @@ def inspect_file(
...
-        elif line_strip == "OPTIMIZE" and fmu_include:
-            section = "optimize/"
-            if not os.path.exists("%s/include/%s" % (packing_path, section)):
-                os.makedirs("%s/include/%s" % (packing_path, section))
+        elif line_strip == "OPTIMIZE" and fmu_include:  # type: ignore
+            section = "optimize/"  # type: ignore
+            if not os.path.exists(
+                "%s/include/%s" % (packing_path, section)  # type: ignore
+            ):
+                os.makedirs("%s/include/%s" % (packing_path, section))  # type: ignore
```
jondequinor, line 502: What's going on here? And could this maybe be better handled by ignoring the file in a mypy config?

.github/workflows/subscript.yml
jondequinor: OKay, I see it was added (at the bottom).

src/subscript/pack_sim/pack_sim.py
berland: You've found the black sheep in subscript.. `pack_sim` is in need for a general rewrite yes, (see #363). I struggled a bit with this part but found no better way than ignoring these lines. Ignoring the whole file is perhaps a little bit too permissive?

jondequinor: This is cluttering the history, so I lean towards ignoring the whole file.

berland: Will fix 👍🏻

jondequinor (approved):
Skimmed most parts, looked closely on some. Overall I think this is good after the excessive `# type: ignore` fix.

Issue #361, opened 2021-05-04
[casegen_upcars] Add test triggering set_vug()


berland:
Codecov uncovers  that a big portion of casegen_upcars is not tested. A small/simple test triggering this function should be added.

https://codecov.io/gh/equinor/subscript/src/master/src/subscript/casegen_upcars/model.py#L863

PR #366, opened 2021-05-06
Add option to shift and rotate cell coordinates

alifbe:
Users requested option to specify origin which is by default is 0, 0. In addition, users also required the capability to rotate the cells coordinates.

image


ertomatic:

berland:
Could some tests for this be added? Maybe covered with the additions in the yaml-file, but that only tests with zero numbers.

alifbe:
Hi @berland 

Thank you for the review. I have now added two more test cases. One for the vugs (#361) and another one is for coordinate transformation case.

PR #369, opened 2021-05-10
casegen_upcars: Ensure compatibility with previous configuration file

alifbe:
Previous configuration file (yaml file) doesn't have the parameters for originX, originY and rotation. This will make new casegen_upcars failed to run. This modification put default value of 0.0 for all the three parameters if they don't exist in configuration file. 

In addition, I have also fixed some mistakes in the comment (yaml file).


berland:
This is a good example of why there should be more extensive tests. I would suggest having test-functions that generate different yaml-files programmatically (and dump to a `tmpdir`), and then ensure that the tool runs correctly on all of them. There are surely bugs lurking in different corner cases. Use also the codecov tool to see which parts of the code is not touched by tests at all: 

https://codecov.io/gh/equinor/subscript/tree/master/src/subscript/casegen_upcars

berland:
@alifbe are you able to add some more tests? This should be merged early next week to be in time for the next komodo release.

alifbe:
> @alifbe are you able to add some more tests? This should be merged early next week to be in time for the next komodo release.

@berland  I am not sure if I understand it properly. Do you have any example that I can use as reference?

berland:
Just more of what you already have in `test_casegen_upcars.py`, covering more different use-cases.

alifbe:
> Just more of what you already have in `test_casegen_upcars.py`, covering more different use-cases.

@berland  I have now added a couple more test and increases the code coverage. There are still a bit missing at some places. But I hope this is good enough for June release. It is rather important to have it push to stable komodo to keep compatibility with old yaml file.

src/subscript/casegen_upcars/model.py
```diff
@@ -92,6 +92,8 @@ def __init__(
 
         if isinstance(streak_rect, list) and len(streak_rect) > 0:
             if isinstance(streak_rect[0], list) and len(streak_rect) > 1:
+                print(len(streak_rect))
```
berland, line 95: leftover debugging statements?

alifbe: @berland Good catch. Yes, it is. I will remove it now.

berland (approved):
Looks good!

PR #371, opened 2021-05-14
Remove region export in prtvol2csv

berland:
Based on top of typehinting branch, look only at the second commit when reviewing.

### DIFF
--- a/src/subscript/config_jobs/PRTVOL2CSV
+++ b/src/subscript/config_jobs/PRTVOL2CSV
@@ -1,14 +1,10 @@
-EXECUTABLE prtvol2csv 
+EXECUTABLE prtvol2csv
 
 
 DEFAULT <DIR> share/results/volumes
 DEFAULT <OUTPUTFILENAME> simulator_volume_fipnum.csv
 
--- Warning: simulator_volume_region.csv is deprecated and
--- will be removed later
-DEFAULT <REGIONOUTPUTFILENAME> simulator_volume_region.csv
-
-DEFAULT <REGIONS> "" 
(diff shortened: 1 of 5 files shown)


berland:
The region export is removed to facilitate refactoring in `fipmapper` which is due to be moved into `fmu.tools`. `fipmapper` will be extended to automatically determine which fipnum-region-zones are comparable in volumes based on a fipmap.

eivindsm (commented):
This PR looks good to me, but as I am not familiar with type hinting in the first place I propose to ask for a review from an IT professional, eg someone from SIB.

Question: are you proposing type hinting to be added to the style recommendation for the repo? This will increase the threshold for contributors, and should probably be aligned a bit before introduced. I'm not opposing type hinting, but worry that the threshold for contributing code becomes to demanding at some point.

berland:
> Question: are you proposing type hinting to be added to the style recommendation for the repo? This will increase the threshold for contributors, and should probably be aligned a bit before introduced. I'm not opposing type hinting, but worry that the threshold for contributing code becomes to demanding at some point.

Not added as a recommendation yet, and definitely not as a a requirement.

berland:
> This PR looks good to me, but as I am not familiar with type hinting in the first place I propose to ask for a review from an IT professional, eg someone from SIB.

The type hinting is reviewed in #362 . Your review of this PR only needs to encompass the second commit ;)

eivindsm (approved):
Second commit looks good

berland:
Converted to draft to ensure #362 is reviewed and merged first.

PR #373, opened 2021-05-18
[sunsch] Adding support for clocktimes

vkip:
Adding support for clocktimes in sunsch by using a local version of time_vector.py (the version from current opm-common/master, to be generally available with the next OPM release).

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -43,3 +43,6 @@ ignore_directives=argparse,automodule
 
 [mypy-*]
 ignore_missing_imports = True
+
+[isort]
+profile=black


ertomatic:

src/subscript/sunsch/time_vector.py
```diff
@@ -0,0 +1,409 @@
+import datetime
```
berland, line 1: Add a comment at the top that this file is due to be removed by November 2021, when OPM 2021.10 is out.

tests/test_sunsch.py
```diff
@@ -344,9 +344,8 @@ def test_days_float(readonly_datadir):
     # The TimeVector object has the "correct" date including time,
     # being 0.1 days after 2020-1-11
     assert datetime.datetime(2020, 1, 11, 2, 24, 0) in sch.dates
-    # However, the clocktime is not included when the TimeVector
-    # object is stringified:
-    assert "11 'JAN' 2020/" in str(sch)
+    # Now also the clocktime is stringified (using updated TimeVector)
```
berland, line 347: Not needed comment, this is git history! :)

tests/test_sunsch.py
```diff
@@ -344,9 +344,8 @@ def test_days_float(readonly_datadir):
     # The TimeVector object has the "correct" date including time,
```
berland, line 344: Maybe remove this comment as well, as it does not make much sense any longer.

tests/test_sunsch.py
```diff
@@ -374,8 +373,8 @@ def test_starttime(readonly_datadir):
         "insert": [{"filename": "foo1.sch", "days": 10}],
     }
     sch = sunsch.process_sch_config(sunschconf)
-    # Dates are rounded down, clock-times are not supported
```
berland, line 377: can you check the assert in line 368 above, that 00:00 is not included? (unless we want it included?)

vkip: It will actually be included - updated assert accordingly.

berland: And there are no side-effects in existing usage of `sunsch` when "00:00:00" is included?

tests/test_sunsch.py
```diff
@@ -355,8 +354,8 @@ def test_days_float(readonly_datadir):
     }
     sch = sunsch.process_sch_config(sunschconf)
     assert datetime.datetime(2020, 1, 11, 21, 36, 0) in sch.dates
-    # Rounding is downwards:
-    assert "11 'JAN' 2020/" in str(sch)
+    # TimeVector now support clocktimes
+    assert "11 'JAN' 2020 21:36" in str(sch)
```
berland, line 355: The Eclipse manual does not exemplify that 21:36 is an ok substitute for 21:36:00. Can you confirm that Eclipse will not choke on it?

berland:
Should this PR get ready and tested in bleeding before being released?

berland:
Merge window for komodo 2021.07 closes Thursday. Will this get there?

PR #374, opened 2021-05-19
Skip binary files in eclcompress

### LINKED ISSUE #243: [eclcompress] detect if file is binary grdecl and skip
When export from RMS is binary grdecl, eclcompress on that file should be skipped. 

This is a case since binary grdecl is not supported for all Eclipse keywords, and "mixed-exports" will occur.

See if it is possible for eclcompress to quickly detect if file is binary an just omit any actions. Perhaps like this: https://stackoverflow.com/a/30273352

berland:
Resolves #243


src/subscript/eclcompress/eclcompress.py
```diff
@@ -145,6 +152,20 @@ def eclcompress(files, keeporiginal=False, dryrun=False, eclkw_regexp=None):
...
+
+    This uses the same heuristic as the Linux `file` command. Only the first
+    1024 bytes are looked at.
+
+    Args:
+        filename: File to check
+    """
+    # https://stackoverflow.com/questions/898669/how-can-i-detect-if-a-file-is-binary-non-text-in-python  # noqa
+    textchars = bytearray({7, 8, 9, 10, 12, 13, 27} | set(range(0x20, 0x100)) - {0x7F})
+    return bool(open(filename, "rb").read(1024).translate(None, textchars))
```
anders-kiaer, line 166:
```suggestion
    with open(filename, "rb") as fh:
        return bool(fh.read(1024).translate(None, textchars))
```
...in order to properly close the file handle?

anders-kiaer (approved):

src/subscript/eclcompress/eclcompress.py
```diff
@@ -75,6 +76,11 @@ def eclcompress(
     totalsavings = 0
 
     for filename in files:
+
+        if file_is_binary(filename):
+            logger.info("Skipped %s, not text file", filename)
+            continue
```
jondequinor, line 82: I think guessing what a file is based on its content is a bad solution, must be absolute last resort.

- What generates these files?
- Why are binary files ignored?
- Who wants these files compressed and for what purpose?

berland: I agree in general, but evaluate it to be worth it in this situation.

* The upside in disk storage savings is significant. It can be gigabytes pr. realization.
* Letting users inject only `FORWARD_MODEL ECLCOMPRESS` in the ert config makes it very simple for users, but requires binary sniffing.

It is already possible for users to be explicit in which files to compress, but I think the upside in actually getting compression done outweighs the downsides from implicitness, therefore the recommended configuration in documentation is do it the simple way.

Binary files and ASCII-files (that should be compressed) are both outputted by RMS, and the file format is a choice made by the geomodeller when setting up the file exports in RMS.

The better solution is have RMS do run-length encoding of ascii export by default, but that feature is not there (yet?)

jondequinor: Thanks for explaining, and that makes total sense.

jondequinor (approved):
LGTM

PR #377, opened 2021-05-26
[prtvol2csv] Use fipmapper from fmu.tools

berland:
Merge this when there has been a fmu.tools with fipmapper in a released version for a reasonable time.


berland:
Ready to be merged. The tool `rmsecl_volumetrics` already depend on fipmapper from fmu-tools, and fmu-tools is also now in pypi.

PR #378, opened 2021-05-27
[fmuobs] Set uniform loglevel for submodules

berland:
Not sure if this is the way to do it, but have not found a solution that automates this through interitance of the log level for already initialized loggers.

### DIFF
--- a/src/subscript/fmuobs/fmuobs.py
+++ b/src/subscript/fmuobs/fmuobs.py
@@ -308,13 +308,16 @@ def fmuobs(
 ):
     # pylint: disable=too-many-arguments
     """Alternative to main() with named arguments"""
-    if verbose:
+    if verbose or debug:
         if __MAGIC_STDOUT__ in (csv, yml, ertobs):
-            raise SystemExit("Don't use verbose mode when writing to stdout")
-        logger.setLevel(logging.INFO)
-
-    if debug:
-        logger.setLevel(logging.DEBUG)
+            raise SystemExit("Don't use verbose/debug when writing to stdout")
+        loglevel = logging.INFO
+        if debug:
+            loglevel = logging.DEBUG
+        logger.setLevel(loglevel)
+        getLogger("subscript.fmuobs.parsers").setLevel(loglevel)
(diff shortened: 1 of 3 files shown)


src/subscript/fmuobs/fmuobs.py
```diff
@@ -312,9 +312,14 @@ def fmuobs(
...
         logger.setLevel(logging.INFO)
-
+        getLogger("subscript.fmuobs.parsers").setLevel(logging.INFO)
+        getLogger("subscript.fmuobs.writers").setLevel(logging.INFO)
+        getLogger("subscript.fmuobs.util").setLevel(logging.INFO)
     if debug:
         logger.setLevel(logging.DEBUG)
+        getLogger("subscript.fmuobs.parsers").setLevel(logging.DEBUG)
+        getLogger("subscript.fmuobs.writers").setLevel(logging.DEBUG)
+        getLogger("subscript.fmuobs.util").setLevel(logging.DEBUG)
```
jcrivenaes, line 323: I think you can make this shorter by avoiding repetition:

```
loglevel = logging.INFO
if debug:
     loglevel = logging.DEBUG

logger.setLevel(loglevel)
getLogger("subscript.fmuobs.parsers").setLevel(loglevel)
...
```

jcrivenaes:
Not sure exactly what you want to achieve, but as idea(?) I have set fmu-dataio to inherit logging level by using a trick in the __init__.py file: https://github.com/equinor/fmu-dataio/blob/main/src/fmu/dataio/__init__.py

berland:
> Not sure exactly what you want to achieve, but as idea(?) I have set fmu-dataio to inherit logging level by using a trick in the **init**.py file: https://github.com/equinor/fmu-dataio/blob/main/src/fmu/dataio/__init__.py

Part of the problem is that the loggers are already initialized when the `--verbose` or `--debug` command line options are parsed.

jcrivenaes (approved):
I think this is OK as it is just logging (messages) behaviour.

src/subscript/fmuobs/fmuobs.py
berland: fixed

PR #384, opened 2021-05-31
[interp_relperm] Support family 2 input and pyscal-xlsx input

berland:
Refactoring and enhancements for interp_relperm to support family 2 input and also pyscal xlsx input.


eivindsm (commented):
LGTM, except for the DESCRIPTION which is now outdated with respect to support for family ii and the possibility for initializing from spreadsheet.

eivindsm (approved):
LGTM

PR #385, opened 2021-06-01
[interp_relperm] Drop support for missing low/high

berland:
And some refactoring to later allow for flexible input.

### DIFF
--- a/src/subscript/interp_relperm/interp_relperm.py
+++ b/src/subscript/interp_relperm/interp_relperm.py
@@ -262,10 +262,20 @@ def make_wateroilgas(dframe: pd.DataFrame, delta_s: float) -> pyscal.WaterOilGas
     """
     wog = pyscal.WaterOilGas(swl=dframe["SW"].min(), h=delta_s)
     wog.wateroil.add_fromtable(
-        dframe[["SW", "KRW", "KROW", "PCOW"]].dropna().reset_index()
+        dframe[["SW", "KRW", "KROW", "PCOW"]].dropna().reset_index(),
+        # For pyscal <= 0.7.7:
+        swcolname="SW",
+        krwcolname="KRW",
+        krowcolname="KROW",
+        pccolname="PCOW",
     )
     wog.gasoil.add_fromtable(
-        dframe[["SG", "KRG", "KROG", "PCOG"]].dropna().reset_index()
+        dframe[["SG", "KRG", "KROG", "PCOG"]].dropna().reset_index(),
+        # For pyscal <= 0.7.7:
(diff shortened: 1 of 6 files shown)


src/subscript/interp_relperm/interp_relperm.py
```diff
@@ -156,23 +152,9 @@ def _is_valid_interpolator(interp: dict):
...
-        pass
-
-    try:
-        if schema["high"]:
-            valid = True
-    except (KeyError, ValueError, TypeError):
-        pass
-
-    return valid
+    return "base" in schema and "low" in schema and "high" in schema
```
eivindsm, line 157: I guess the need for the entire function disappeared, since base, low and high now can be set to required in the schema.

eivindsm: The same goes for the function below ..

eivindsm (commented):
This looks really good and is much simpler than the original commit!

src/subscript/interp_relperm/interp_relperm.py
berland: The validator for low-base-high was not even active! I will keep it in, since I need it in the next PR. 

family validator was not active either, now fixed. Did not find an enumeration class in configsuite, so I guess it should not be dropped?

eivindsm: Ups, it should have been active. Well spotted!

berland: Yes, but https://github.com/equinor/subscript/pull/385/commits/c48cbfd3f46c4ddff71c4049dadf0b4da7d4ba02#diff-1cb24c63cb076e87801721182f5481b715e6763706cc6e1dd266b9f0e1ad6ce4R171 does the trick. 

Will add tests for this in next PR.

berland: (you might be right, before this PR it might have error gracefully after schema validation if low/high/base was missing).

berland: Fixes and tests are now applied to this PR.

eivindsm (approved):
PR looks good and I approved, but I still think the need for the _is_valid_table_entries decorator is obsolete as you should be able to set high low and base as required in the schema.

berland:
> PR looks good and I approved, but I still think the need for the _is_valid_table_entries decorator is obsolete as you should be able to set high low and base as required in the schema.

(and I agree, but it is needed again in next PR when low-base-high is not required if a xlsx file is supplied)

PR #389, opened 2021-06-04
Fix issues in pack_sim

berland:
* Avoid global variables. Allow parallel testing.
* Use pathlib.Path more extensively
* Use subscript logger instead of print
* Add enforced mypy
* Raise ValueError if DATA-file in packing path already exists (this is a change in behaviour for the script, towards predictability)


wouterjdb (approved):
Nice clean-up work there @berland 🎉

PR #394, opened 2021-06-15
Make interp_relperm robust to future pyscal versions

berland:
Without this fix, subscript is only compatible with pyscal <= v0.7.7

Problem is only visible when run towards master branch of pyscal (that means locally or in komodo bleeding).


eivindsm (approved):
LGTM

PR #397, opened 2021-06-17
[welltest_dpds] update script to include water phase

mareco701:
Update welltest_dpds script to include calculation for water phase as well.

### DIFF
--- a/src/subscript/welltest_dpds/welltest_dpds.py
+++ b/src/subscript/welltest_dpds/welltest_dpds.py
@@ -29,8 +29,8 @@ Outputs files according to naming convention outputdirectory/fname_outfilesuffix
   * dpdspt_lag2_genobs_suffix_bunr; if --genobs_resultfile is invoked
   * wbhp_genobs_suffix_bunr; if --gen_obs_result_file is invoked
 
-gen_obs_result_file is to be invoked to generate necessary files to be sued with GENERAL_OBSERVATION in ERT.
-Note: the outfilesuffix argument is then used to pass on the RESTART/report step number. 
+gen_obs_result_file is to generate files to be used with GENERAL_OBSERVATION in ERT.
+Note: the outfilesuffix argument is then used to pass on the RESTART/report step number.
 
 """
 
@@ -61,7 +61,8 @@ EXAMPLES = """
  FORWARD_MODEL  WELLTEST_DPDS(<ECLBASE>, <WELLNAME>=OP_1, <PHASE>=GAS,
     <BUILDUP_NR>=1, <OUTPUTDIRECTORY>=dst, <OUTFILESSUFFIX>=OP_1_1)
 
-Then GEN_DATA DPDT_SIM  INPUT_FORMAT:ASCII REPORT_STEPS:1 RESULTS_FILE:dpdspt_lag2_genobs_<WELLNAME>_%d_<BUILDUP_NR>
+Then GEN_DATA DPDT_SIM  INPUT_FORMAT:ASCII REPORT_STEPS:1
+RESULTS_FILE:dpdspt_lag2_genobs_<WELLNAME>_%d_<BUILDUP_NR>
 """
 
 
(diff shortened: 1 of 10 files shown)


ertomatic:

eivindsm (commented):
PR looks good. With regard previous complaints about option --genobs_resultfile being difficult to understand; should we consider to improve the parser help attribute? 

Original:
File with welltest results used to define time steps, typically exported from Saphir. If present, additional files to be used with GENERAL_OBSERVATION in ERT are produced

Suggestion:
Option to trigger export of result files compatible with ERTs GENERAL_OBSERVATION file format to be used in history matching. Expected argument is a file with welltest results used to define the time steps to be reported. This file is typically exported from Saphir.

src/subscript/welltest_dpds/welltest_dpds.py
```diff
@@ -18,16 +18,20 @@
...
 
 Outputs files according to naming convention outputdirectory/fname_outfilesuffix.csv:
   * dpdspt_lag1; cumtime and superpositioned time derivative of pressure lag 1
   * dpdspt_lag2; cumtime and superpositioned time derivative of pressure lag 2
   * spt; superpositioned time
   * welltest; unified csv file with vectors: cumtime, wbhp, wopr, wgpr, wwpr
-  * dpdspt_lag1_gendobs_suffix_bunr; if --genobs_resultfile is invoked
-  * dpdspt_lag2_gendobs_suffix_burn; if --genobs_resultfile is invoked
+  * dpdspt_lag1_genobs_suffix_bunr; if --genobs_resultfile is invoked
+  * dpdspt_lag2_genobs_suffix_bunr; if --genobs_resultfile is invoked
```
berland, line 29: bunr or burn? (does it mean something?)

mareco701: it stands for build up number (bunr)

src/subscript/welltest_dpds/welltest_dpds.py
```diff
@@ -451,16 +457,24 @@ def main():
...
         wopr = summary_vec(summary, "WOPR:" + well_name, required=False)
         wgpr = summary_vec(summary, "WGPR:" + well_name)
+        wwpr = summary_vec(summary, "WWPR:" + well_name, required=False)
+    else:
+        wopr = summary_vec(summary, "WOPR:" + well_name, required=False)
+        wgpr = summary_vec(summary, "WGPR:" + well_name, required=False)
+        wwpr = summary_vec(summary, "WWPR:" + well_name)
 
-    wwpr = summary_vec(summary, "WWPR:" + well_name, required=False)
+    #    wwpr = summary_vec(summary, "WWPR:" + well_name, required=False)
```
berland, line 470: (delete line)

tests/test_welltest_dpds.py
```diff
@@ -55,7 +55,13 @@ def test_main(tmpdir, mocker):
...
     assert Path("welltest_output.csv").exists()
     Path("welltest_output.csv").unlink()
-
+    # test --phase
+    mocker.patch(
+        "sys.argv", ["welltest_dpds", DATAFILEPATH, "55_33-1", "--phase", "WATER"]
+    )
+    welltest_dpds.main()
+    assert Path("welltest_output.csv").exists()
+    Path("welltest_output.csv").unlink()
```
berland, line 64: Is there something in the csv that we can assert on that proves that the WATER argument was processed, and processed correctly?

mareco701: unfortunately there is nothing in the output files that specifies for which phase it has been calculated. So the test is actually only checking  that water can be an argument and that it will generate a csv file. The test data used have the same start/end date for the build up whatever the phase.

mareco701: However we could check the superposition time calculation which is using the rate (thus will change depending on the phase)...

src/subscript/welltest_dpds/welltest_dpds.py
mareco701: yes sorry I forgot

berland:
FYI: If merged within Thursday 24, this will be part of the komodo 2021.07 release.

mareco701:
It would be nice, Someone needs to approve before we can merge.

eivindsm (approved):
LGTM, but I suggest that we also update the description included in the ert doc with respect to  option --gen_obs_result_file

PR #398, opened 2021-06-18
Rmsecl volumetrics

berland:
Finally a tool for comparing Eclipse and RMS volumetrics. 

Not sure if the naming is good.

This might primarily be a tool that is used in Webviz, but should possibly be a command line utility also.

Suggest to release this in komodo 2021.07 but considering limited news coverage before it hits Webviz.


berland:
Depends on https://github.com/equinor/fmu-tools/pull/143

docs/scripts/rmsecl_volumetrics.rst
```diff
@@ -0,0 +1,12 @@
+
+RMSECL_VOLUMETRICS
```
jcrivenaes, line 2: I think we need a more concrete example in the documentation

berland: Now added.

jcrivenaes: Excellent!

jcrivenaes (approved):
LGTM!

PR #401, opened 2021-06-24
fmu-tools is on pypi

berland:


jcrivenaes (approved):
Looks good but there is a need for a GH config that uploads to pypi on releases.

PR #403, opened 2021-06-29
Make contrib instructions csh-compatible

berland:


asnyv:
Long time since I used `csh`, but think you also need a `.csh` after `source venv-subscript/bin/activate` when using `csh` (at least in Equinor)?

berland:
> Long time since I used `csh`, but think you also need a `.csh` after `source venv-subscript/bin/activate` when using `csh` (at least in Equinor)?

Good catch!

asnyv (approved):

PR #404, opened 2021-06-29
Use isort for import sorting

berland:


.github/workflows/subscript.yml
```diff
@@ -41,6 +41,8 @@ jobs:
 
       - name: Check code style
         run: |
+          pip install isort
+          isort --profile black src tests
           pip install black
```
anders-kiaer, line 46: Since `black` and `isort` is part of `tests` requirements, should `subscript` installation be done prior to "Check code style" such that you can simply use the `tests` requirements installed from there (reducing risk of GitHub workflow and that file getting out of sync).

anders-kiaer (approved):
One comment, otherwise LGTM.

.github/workflows/subscript.yml
berland: Pushed code style checking to below installation and simplified.

PR #405, opened 2021-07-01
Tighten typing in eclcompress/API

berland:


src/subscript/eclcompress/eclcompress.py
```diff
@@ -190,20 +188,22 @@ def acceptedvalue(valuestring: str) -> bool:
...
-        keywordsets (list): List of 2-tuples, (start, end) indices in
+        keywordsets: List of 2-tuples, (start, end) indices in
             line number in the deck, referring to individual sections
             of distinct keywords.
-        filelines (list): list of lines (strings) from Eclipse deck, cleaned.
+        filelines: list of lines from Eclipse deck, cleaned.
 
     Returns:
-        list:  List of strings to be used as a replacement Eclipse deck
+        Strings to be used as a replacement Eclipse deck
```
asnyv, line 206: In another method you have switched to e.g. `List of strings [...]` and `List of 2-tuples [...]` , so maybe change this for consistency, even though the type hinting is even more descriptive?
```suggestion
        List of strings to be used as a replacement Eclipse deck
```

asnyv (changes requested):

src/subscript/eclcompress/eclcompress.py
berland: The intention was not to say "list of something" when the type hinting already says the same (and it is included next to the text in sphinx-generated docs), only a textual description, and then "strings" in plural is the textual description of "List[str]"

berland: Consistency now attempted.

asnyv (approved):
⛱

Issue #234, opened 2020-11-06
[fmu_copy_revision] allow for listing of revisions following new proposed namestandard


jcrivenaes:
Currently fmu_copy_revision (unless `--all` option is used) will look for folders named r000 til r999. 

A new versioning system is proposed in FMU, and listing those without the need for `--all` shall be possible

Also consider to simplify the folders by skipping one level, e.g. go from:

`users/nnn/20.1.0/20.1.0_20200901`

to

`users/nnn/20.1.0_20200901`

PR #407, opened 2021-08-03
Rewrite fmu copy revision

jcrivenaes:
New version of fmu_copy_revision, fully rewritten i Python
* Will handle revisons of type 19.1.0 and r003 by default, issue #234 
* Use in general xargs to do use multiple threads. This speeds up copying usually by a factor 4 for large projects on RGS
* Also possible to run this script with command line opions only
* More options; the default is now that empty folders (at target; after filtering) will not be copied


ertomatic:

tests/test_fmu_copy_revision.py
```diff
@@ -0,0 +1,133 @@
...
+]
+
+
+@pytest.fixture(name="datatree", scope="session", autouse=True)
+def fixture_datatree(tmp_path_factory):
+    """Create a tmp folder structure for testing."""
+    tmppath = tmp_path_factory.mktemp("data")
+    for top in TOPLEVELS:
+        (tmppath / top).mkdir(parents=True, exist_ok=True)
+        for fil in FILESTRUCTURE:
```
berland, line 36: fil -> file

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,656 @@
...
+from multiprocessing import cpu_count
+
+
+import subscript
+from subscript import __version__
+
+logger = subscript.getLogger(__name__)
+
+DESCRIPTION = """This is a simple interactive script for copying a FMU revision folder
+with options of:
```
berland, line 24: Suggest: "with options of" -> "with features:"

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,656 @@
...
+    1. Selective copy, i.e. avoid data that can be regenerated
+    2. Speed up copying by multithreading
+    3. Retain correct file dates and user permissions
+
+Usage:
+    fmu_copy_revision  (for menu based input)
+
+     * or *
+
+    fmu_copy_revision --source 21.0.0 --target whatever --option 3 --threads 6 -cleanup
```
berland, line 34: Should we rename `option` here to `mode` or `copymode` or something?

berland: double dash in `cleanup`?

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,656 @@
...
+
+     * or *
+
+    fmu_copy_revision --source 21.0.0  (...other options are automated)
```
berland, line 38: automated -> defaulted?

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,656 @@
...
+            args = sys.argv[1:]
+
+        parser = get_parser()
+        self.args = parser.parse_args(args)
+
+    def check_folders(self):
+        """Check if potential fmu folders are present or list all if --all."""
+
+        current = pathlib.Path(".")
+        folders = [fil for fil in current.iterdir() if fil.is_dir()]
```
berland, line 321: fil -> file

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,656 @@
...
+
+        if result:
+            result = sorted(result)
+
+        if not result:
+            print(
+                "No valid folders to list. Are you in the correct folder "
+                "above your revisions? Or consider --all option to list all folders."
+            )
+            sys.exit()
```
berland, line 339: Exception rather than a system exit?

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,656 @@
...
+            raise ValueError("Input folder does not exist!")
+
+        today = time.strftime("%Y%m%d")
+        user = getpass.getuser()
+        logger.info("User and today: %s %s", user, today)
+
+        userpath = sourcepath.parent / "users"
+        if not userpath.exists():
+            userpath.mkdir(parents=False, exist_ok=True)
+            print("Made folder: users")
```
berland, line 383: logger.info?

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,656 @@
...
+        if not result:
+            print(
+                "No valid folders to list. Are you in the correct folder "
+                "above your revisions? Or consider --all option to list all folders."
+            )
+            sys.exit()
+
+        self.folders = result
+
+    def menu1(self):
```
berland, line 343: Can this function be renamed to something more descriptive? Something that more clearly separates what is `menu1` and `menu2`

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,656 @@
...
+        default = "4"
+        self.option = int(input(f"Choose (default is {default}): ") or default)
+
+        if self.option == 9:
+            ffile = input("Choose rsync filter file: ")
+            with open(ffile, "r") as stream:
+                self.filter = stream.read()
+
+    def define_filterpattern(self):
+        """Define filterpattern pattern based on menu choice."""
```
berland, line 488: (not only menu choice, also command line I suppose)

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,656 @@
...
+            stream.write(self.filter)
+
+        with open(dirfilterpatternname, "w") as stream:
+            stream.write(self.dirfilter)
+
+        logger.info("FILTER:\n %s", self.filter)
+
+        self.nthreads = self.args.threads
+        if self.nthreads == "99":
+            # as default, leave one CPUree for other use
```
berland, line 551: CPUree ?

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,656 @@
...
+            rsyncargs += " --dry-run -v"
+
+        if self.args.verbosity:
+            rsyncargs += " -v"
+
+        # execute rsync, some explanations:
+        # filterpatternname: is the file to do per file filtering
+        # rsyncargs: see above
+        # self.keepfolders will be:
+        #   - 0 to delete all folders tha tempty after copy/filtering
```
berland, line 569: typos

tests/test_fmu_copy_revision.py
jcrivenaes: OK (just code variable though)

src/subscript/fmu_copy_revision/fmu_copy_revision.py
jcrivenaes: > Should we rename `option` here to `mode` or `copymode` or something?

Or perhaps `profile`?

src/subscript/fmu_copy_revision/fmu_copy_revision.py
jcrivenaes: Will prefer sys.exit() since this is a end-point user script. There is no way the user could test on the Exception and do something else. A sys.exit() will give a clean exit while Exception will give a noisy message.

src/subscript/fmu_copy_revision/fmu_copy_revision.py
jcrivenaes: Good idea

src/subscript/fmu_copy_revision/fmu_copy_revision.py
jcrivenaes: One `CPU free ` ...

berland:
Can tests for the interactive menu also be written? Similar to https://github.com/equinor/subscript/blob/master/tests/test_sw_model_utilities.py

berland:
I assume we keep `fmu_copy_revision_old` for some time to have a solution for users when the new tool misbehaves. But there should potentially be a deprecation warning added to the output of the legacy script?

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,655 @@
+#!/usr/bin/env python
```
berland, line 1: don't think we need this

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,655 @@
...
+import sys
+from os.path import join
+import argparse
+import pathlib
+import tempfile
+import time
+import getpass
+import subprocess
+import shutil
+from multiprocessing import cpu_count
```
berland, line 15: Can you do `pip install isort; isort fmu_copy_revision.py`  ?

jcrivenaes: Done!

src/subscript/fmu_copy_revision/fmu_copy_revision.py
jcrivenaes: OK I will remove

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,654 @@
...
+DIRFILTERFILE="$7"
+
+PWD=$(pwd)
+
+start=`date +%s.%N`
+
+cd $SRCDIR
+
+echo " ** Target folder is $DESTDIR"
+echo " ** Filter file is $FILTERFILE"
```
berland, line 198: This filter file is removed if everything works well, so this echo should perhaps be muted (potentially printed as part of a potential error message, to aid debugging)

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,654 @@
...
+        formatter_class=argparse.RawTextHelpFormatter,
+        usage=usetext,
+    )
+
+    parser.add_argument("--dryrun", action="store_true", help="Run dry run for testing")
+    parser.add_argument(
+        "--all, -all, -a", action="store_true", dest="all", help="List all folders"
+    )
+    parser.add_argument(
+        "--verbosity, -v",
```
berland, line 244: (this kind of option is named "verbose" in most other tools in subscript)

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,654 @@
...
+
+        sourcekbytes = fsize // 1024
+        if sourcekbytes > freekbytes:
+            print("Not enough space left for copying! STOP!")
+            sys.exit()
+
+        time.sleep(1)
+
+    def show_possible_profiles_copy(self):
+        """Show a menu for possuble profiles for copy/rsync."""
```
berland, line 473: possuble -> possible

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,654 @@
...
+
+        fsize = _get_size(self.source)
+        print(f"\n  Size of existing revision is: {_filesize(fsize)}\n")
+
+        sourcekbytes = fsize // 1024
+        if sourcekbytes > freekbytes:
+            print("Not enough space left for copying! STOP!")
+            sys.exit()
+
+        time.sleep(1)
```
berland, line 470: Needed?

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,654 @@
...
+        print("Choices:\n")
+
+        default = 0
+        for inum, res in enumerate(self.folders):
+            print(f"{inum + 1:4d}:  {res}")
+            default = inum + 1  # take last as default
+        try:
+            select = int(input(f"\nChoose number, default is {default}: ") or default)
+        except ValueError:
+            print("Selection is not a number")
```
berland, line 352: missing sys.exit() here?

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,654 @@
...
+        "--profile",
+        dest="profile",
+        type=str,
+        default="3",
+        help="profile for copy profile to use, default is 3",
+    )
+    parser.add_argument(
+        "--threads",
+        dest="threads",
+        type=str,
```
berland, line 274: integer? If a  string is passed, it says "Do copying using foobar CPU threads, wait..." (everything seems to work, but an error should probably be issued).

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,654 @@
...
+            stream.write(self.dirfilter)
+
+        logger.info("FILTER:\n %s", self.filter)
+
+        self.nthreads = self.args.threads
+        if self.nthreads == "99":
+            # as default, leave one CPU free for other use
+            self.nthreads = cpu_count() - 1 if cpu_count() > 1 else 1
+
+        print(f"Do copying using {self.nthreads} CPU threads, wait...")
```
berland, line 552: Do -> Doing/Starting/Performing?

berland:
If the target directory already exists, it will currently silently overwrite. Is that intentional, or should we error hard and require the `--cleanup` option?

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,654 @@
...
+            runner.construct_default_target()
+            proposal = runner.default_target
+        else:
+            proposal = pathlib.Path(runner.args.target)
+        runner.construct_target(proposal)
+        runner.check_disk_space()
+        runner.define_filterpattern()
+        print(
+            f"Using source <{runner.source}>, target <{runner.target}> with "
+            f"option <{runner.profile}> ..."
```
berland, line 648: option -> profile

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,654 @@
...
+        help="If present, skip estimation of current revision size.",
+    )
+    parser.add_argument("--source", dest="source", type=str, help="Add source folder")
+    parser.add_argument("--target", dest="target", type=str, help="Add target folder")
+    parser.add_argument(
+        "--profile",
+        dest="profile",
+        type=str,
+        default="3",
+        help="profile for copy profile to use, default is 3",
```
berland, line 269: Is the default 3 or 4?

src/subscript/fmu_copy_revision/fmu_copy_revision.py
jcrivenaes: True, it can been seen in the logger if wanted

src/subscript/fmu_copy_revision/fmu_copy_revision.py
jcrivenaes: OK, I add --verbose as first (--verbosity will still work also, as well as -v)

src/subscript/fmu_copy_revision/fmu_copy_revision.py
jcrivenaes: Not really, but the menu comes very fast so that disk space info disappears if a small xterm window. Think I will keep it

src/subscript/fmu_copy_revision/fmu_copy_revision.py
jcrivenaes: Reformulated to "Doing copy using..."

src/subscript/fmu_copy_revision/fmu_copy_revision.py
jcrivenaes: Corrected to 4

jcrivenaes:
> If the target directory already exists, it will currently silently overwrite. Is that intentional, or should we error hard and require the `--cleanup` option?

Actually rsync can do an update, but not sure how safe that will be (in particular of one changes the profile). So, agree with you. I set the cleanup option to be defaulted to true instead.

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -241,16 +240,17 @@ def get_parser() -> argparse.ArgumentParser:
...
     parser.add_argument(
-        "--verbosity, -v",
+        "--verbose, --verbosity, -v",
         action="store_true",
         dest="verbosity",
         help="Enable logging (messages) for debugging",
     )
     parser.add_argument(
         "--cleanup",
         action="store_true",
```
berland, line 250: This probably does not work if the default is True.

Has this been True by default in the legacy version? Otherwise I would prefer this to be False by default and error hard if target exists.

jcrivenaes:
`--cleanup` is a new option. Earlier it did a merge but will guess most users avoided it may either renaming the output or manually deleting the folder first (it happens only automatically if copying is done the same day). I propose to add the option to `--merge` which is defaulted to false. Then the behaviour should be _if target already exists_:
`fmu_copy_revision`: will fail hard but propose cleanup or merge for the user
`fmu_copy_revision --cleanup`: Will remove previous target first
`fmu_copy_revision --merge`: Will use rsync merging
`--cleanup` and `--merge` cannot coexist on the same command line. 
Agree?

berland:
> Agree?

Yes!

src/subscript/fmu_copy_revision/fmu_copy_revision.py
```diff
@@ -0,0 +1,677 @@
...
+
+        if process.returncode != 0:
+            print(f"Process returncode: {process.returncode}")
+            print(process.stderr.decode())
+        stdout = process.stdout.decode().splitlines()
+
+        output = "\n".join(stdout[0:-2])
+        print(output)
+        timing = float(stdout[-1])
+        timing = time.strftime("%H hours %M minutes %S seconds", time.gmtime(timing))
```
berland, line 629: (this is nicely done with `humanize.precisedelta(sometimedelta)` but `humanize` is not in komodo)

berland (approved):
LGTM!

PR #411, opened 2021-08-12
Fix bug from inplace value substitution in ecldiff2roff

berland:


rnyb (approved):

src/subscript/ecldiff2roff/ecldiff2roff.py
```diff
@@ -204,7 +207,7 @@ def ecldiff2roff_main(
         filename = outputfilebase + sep + diffpropname + ".roff"
 
         logger.info("Writing to file %s", filename)
-        prop1.to_file(filename, name=diffpropname)
+        diffprop.to_file(filename, name=diffpropname)
```
berland, line 210: If the file from `prop1.to_file()`  contained more information than the grid dimensions, this will not give the same roff output  (?)

berland: A sample test confirms that the output from the two ways of exporting (before and after bugfix) yields the same byte output.

jcrivenaes:
FYI Although not directly related to this fix, the style used here will be deprecated i xtgeo soon:
    `ecl_grid = xtgeo.grid3d.Grid().from_file(...`
rather use:
    `ecl_grid = xtgeo.grid_from_file(...`

src/subscript/ecldiff2roff/ecldiff2roff.py
```diff
@@ -191,8 +191,11 @@ def ecldiff2roff_main(
             str(date_pair[0]),
             str(date_pair[1]),
         )
-        # Inplace substraction, prop1 = prop1 - prop2
-        prop1.values -= prop2.values
+
+        diffprop = xtgeo.GridProperty(
+            ncol=ecl_grid.ncol, nrow=ecl_grid.nrow, nlay=ecl_grid.nlay
```
jcrivenaes, line 196: OK, FYI an altaernative and perhaps easier approach would:
`diffprop = prop1.copy()  # makes a new instance`

jcrivenaes (approved):
LGTM

PR #425, opened 2021-09-29
Set rms env after disable_komodo_exec is run

ManInFez:
Using the `env` arg in `subprocess.run` for the rms environment will cause the `disable_komodo_exec` to preserve the rms environment and introduce them in to the `run_external` environment. Therefore we set the envronment between the `disable_komodo_exec` and rms-executable

Partial fix for https://github.com/equinor/komodo-releases/issues/1786

### DIFF
--- a/src/subscript/runrms/runrms.py
+++ b/src/subscript/runrms/runrms.py
@@ -510,17 +510,29 @@ class RunRMS:
 
         rms_exec_env = self._collect_env_settings()
 
-        if shutil.which("disable_komodo_exec"):
-            rms_exec_env["PATH_PREFIX"] = self.setup["roxenv_path"]
-            args_list = ["disable_komodo_exec"] + args_list
+        env_args = ["env"]
+        for key, value in rms_exec_env.items():
+            env_args.append(f"{key}={value}")
 
+        if shutil.which("disable_komodo_exec"):
+            args_list = (
+                [
+                    "env",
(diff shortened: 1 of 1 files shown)


jcrivenaes (approved):
LGTM

src/subscript/runrms/runrms.py
```diff
@@ -510,17 +510,29 @@ def launch_rms(self, empty=False):
...
-        if shutil.which("disable_komodo_exec"):
-            rms_exec_env["PATH_PREFIX"] = self.setup["roxenv_path"]
-            args_list = ["disable_komodo_exec"] + args_list
+        env_args = ["env"]
+        for key, value in rms_exec_env.items():
+            env_args.append(f"{key}={value}")
 
+        if shutil.which("disable_komodo_exec"):
+            args_list = (
+                [
```
sregales-TNO, line 519: I'm pushing it, but it's worth showing off 
```python
        args_list = [
            *(
                [
                    "env",
                    f"PATH_PREFIX={self.setup['roxenv_path']}",
                    "disable_komodo_exec",
                ] 
            if shutil.which("disable_komodo_exec") 
            else []
            ),
            "env", 
            *(f"{key}={value}" for key, value in rms_exec_env.items()),
            *args_list
        ]
```

sregales-TNO (approved):

PR #426, opened 2021-09-29
Add more tests for sector2fluxnum

berland:


berland:
This was an attempt to increase the code coverage, but it more or less failed when I realized that `eclipse` is used in some of the code paths, which are not available in github actions. Still, this PR introduces a new test when run on-prem. 

Code coverage for sector2fluxnum can still be improved very much :)

thomaram:
The eclipse calls were bypassed in the test by supplying pre-defined ECL output as input for testing. Eclipse is just providing the case specific templates for the FLUX file.

tests/test_sector2fluxnum.py
```diff
@@ -1,13 +1,86 @@
...
+    reason="Eclipse must be installed for this test",
+)
+def test_main_notest(tmp_path, mocker):
+    """Test without --test on the command line, requiring
+    Eclipse simulator installed in PATH"""
+    os.chdir(tmp_path)
+
+    input_ECL_CASE = TESTDATA / "TEST.DATA"
     input_OUTPUT_FLUX = "OUT_COARSE.FLUX"
     input_INPUT_DUMPFLUX = os.path.join(TESTDATA, "DUMPFLUX_TEST.DATA")
```
thomaram, line 85: Should we use `os.path.join` for all input, for consistency?

berland: good catch, changed from os.path.join to pathlib path merging, for consistencyy

tests/test_sector2fluxnum.py
```diff
@@ -1,16 +1,89 @@
...
+    sector2fluxnum.main()
+
+    assert Path("USEFLUX_TEST.DATA").exists()
+
+
+@pytest.mark.skipif(
+    not Path("/prog/res/ecl/grid").exists(),
+    reason="Eclipse must be installed for this test",
+)
+def test_main_notest(tmp_path, mocker):
```
thomaram, line 78: Suggest to name function e.g. `test_main_with_ecl_run` to be more explicit.

berland:
Before merge, verify that DUMPFLUX_TEST.INIT  is needed for the added test.

PR #427, opened 2021-09-30
Avoid punching sys.argv in testing, use mocker.patch

berland:


jcrivenaes:
What is the advantage with this?

berland:
Modifying sys.argv directly is supposedly fragile, and I have been instructed to do so earlier. 

https://stackoverflow.com/questions/18668947/how-do-i-set-sys-argv-so-i-can-unit-test-it

jcrivenaes (approved):
LGTM

PR #433, opened 2021-10-11
[sector2fluxnum] Fix test with eclipse

thomaram:
Fixup of test to run with ecl.
- Formatted input strings
- removed args with input files in lieu of files from local ecl run

	modified:   tests/test_sector2fluxnum.py


src/subscript/sector2fluxnum/flux_util.py
```diff
@@ -66,96 +65,6 @@ def filter_region(grid, i, j, k, fipnum, fipnum_kw, combine_operator="intersect"
         )
 
 
-def filter_region_fipnum(grid, fipnum, fipnum_kw, combine_operator="intersect"):
```
berland, line 69: This function is never needed in the script?

thomaram: Probably. The `flux_util.py` is a collection of utility functions, and some of them are superfluous.

PR #436, opened 2021-10-18
Pylinting

berland:


src/subscript/casegen_upcars/model.py
```diff
@@ -340,40 +338,32 @@ def __init__(
...
+        print(f"  dx x dy x dz: {dx} x {dy} x {dz}")
+
+        print(f"  Matrix Element in X-direction: {nMatrixY}")
+        print(f"  Matrix Element in Y-direction: {nMatrixX}")
+        print(f"  Layers: {nz}")
+        print(f"  Shape Factor: {radius_x} (a), {radius_y} (b), {radius_z} (c)")
+        print(f"  Top geometry: {top}")
+        print(f"  Model origin: {origin_x}, {origin_y}")
+        print(f"  Model coordinate rotation: {rotation}°")
+        print(f"  Tilting: {tilt}")
```
jcrivenaes, line 352: Consider using multiline docstrings?
```
import textwrap

x = 44
y = 55


def main():
    multi = f"""
    This is a multiline string
    where things may go wrong for {x}
    but not for {y} and you may
    think
    """

    print(textwrap.dedent(multi))


if __name__ == "__main__":
    main()


"""
```

jcrivenaes (approved):
Many files and changes but it seems to pylinting only so I assume better code style overall

src/subscript/casegen_upcars/model.py
berland: That looked quite cool, but I will leave that for the casegen_upcars developers. :)

berland:
@thomaram , this perturbs many lines in `sector2fluxnum`, but tests pass so we hope it does not cause any disturbance. Feel free to have an extra overlook.

PR #447, opened 2021-10-29
Ensure pytest garbage collection always works

berland:
Interrupted tests typically left directories in /tmp/pytest without user read/write permissions


jcrivenaes (approved):
LGTM

PR #466, opened 2022-01-18
Adding clocktimes to sunsch and enable OPM 2021.10

vkip:


ertomatic:

vkip:
Ah, this is essentially the same as #440, but with a modified time_vector.py to (hopefully) work around the OPM version dep.

berland:
Isn't the `time_vector.py` in OPM 2021.10 sufficient?

berland:
Recommend to require new OPM in this PR and then bump it in komodo accordingly.

vkip:
Actually the time_vector.py in OPM 2021.10 is not sufficient; that version works fine for OPM, but turns out it breaks Eclipse which fails on 0.0000 for the second spec. Now updated with the correct version here at least...

asnyv:
Regarding the other test that will fail when bumping `opm`, the [`test_long_vfp_lines`](https://github.com/equinor/subscript/blob/43b6673186fc059bc993158c4055ae311d107574/tests/test_sunsch.py#L728-L759).
It is caused by [this commit](https://github.com/OPM/opm-common/commit/c44dcb8257ff75da00f291e6485f597bf4d63be7), which reduces the number of "columns" per line for a record of a keyword from 16 to 7, reducing the chance of going above 128 characters. I did a test with to add 20 digits per value in the test we had to make it fail again (7*20 + spaces >> 128), but seems like opm / sunsch somewhere rounds them to 10 digits + the dot as decimal separator, making it impossible to get this error.
Therefore think it is ok to remove the test of long vfp lines, but not sure if that commit fixed the issue for all possible keywords, or if there are keywords where the the values are not rounded. Therefore suggest that we keep the "wrap long lines" code just in case, and rather consider to revisit it later.

asnyv:
> Actually the time_vector.py in OPM 2021.10 is not sufficient; that version works fine for OPM, but turns out it breaks Eclipse which fails on 0.0000 for the second spec. Now updated with the correct version here at least...

How do you mean it fails? The Eclipse documentation says the time should be on format `HH:MM:SSSS` (though default is `00:00:00`, so `HH:MM:SS` is clearly also supported 😅).
Have we tested which formats that are supported by _both_ Eclipse and OPM? Think we should make sure that the output is compatible with both? 
I see that e.g. your tests are on format `HH:MM`, which you and @berland  discussed in this PR https://github.com/equinor/subscript/pull/373, and where it was mentioned that it will make Eclipse choke?

vkip:
Eclipse fails on hh:mm:s.ssss, which the OPM version of time_vector.py produces (my bad, but Flow does not fail). Version here works on both. Ageee we ideally should have a test covering this..

berland:
(the opm pinning in `setup.py` should be removed by this PR)

berland:
> Eclipse fails on hh:mm:s.ssss, which the OPM version of time_vector.py produces (my bad, but Flow does not fail). Version here works on both. Ageee we ideally should have a test covering this..

Ok, there should be a new issue written for subscript then to remove this file when a corresponding fix has been merged upstream.

tests/test_sunsch.py
```diff
@@ -312,8 +311,8 @@ def test_days_float(readonly_datadir):
     }
     sch = sunsch.process_sch_config(sunschconf)
     assert datetime.datetime(2020, 1, 11, 21, 36, 0) in sch.dates
-    # Rounding is downwards:
-    assert "11 'JAN' 2020/" in str(sch)
+    # TimeVector now support clocktimes
+    assert "11 'JAN' 2020 21:36" in str(sch)
```
berland, line 317: issue #440 also tests the `starttime` feature of sunsch. Can  potentially be added, don't remember how important it was. 

Please add something that asserts that microseconds are not leaked into the strings.

vkip:
> Regarding the other test that will fail when bumping `opm`, the [`test_long_vfp_lines`](https://github.com/equinor/subscript/blob/43b6673186fc059bc993158c4055ae311d107574/tests/test_sunsch.py#L728-L759). It is caused by [this commit](https://github.com/OPM/opm-common/commit/c44dcb8257ff75da00f291e6485f597bf4d63be7), which reduces the number of "columns" per line for a record of a keyword from 16 to 7, reducing the chance of going above 128 characters. I did a test with 2021.10 to add 20 digits per value in the test we had to make it fail again (7*20 + spaces >> 128), but seems like opm / sunsch somewhere rounds them to 10 digits + the dot as decimal separator, making it impossible to get this error. Therefore think it is ok to remove the test of long vfp lines, but not sure if that commit fixed the issue for all possible keywords, or if there are keywords where the the values are not rounded. Therefore suggest that we keep the "wrap long lines" code just in case, and rather consider to revisit it later.

Ok - removing test_long_vfp_lines then.

vkip:
> > Eclipse fails on hh:mm:s.ssss, which the OPM version of time_vector.py produces (my bad, but Flow does not fail). Version here works on both. Ageee we ideally should have a test covering this..
> 
> Ok, there should be a new issue written for subscript then to remove this file when a corresponding fix has been merged upstream.

https://github.com/equinor/subscript/issues/467

tests/test_sunsch.py
vkip: Thanks, missed starttime - fixed now, with tests. Also ensured microsecs do not appear where not needed.

src/subscript/sunsch/sunsch.py
```diff
@@ -241,7 +241,9 @@ def get_schema() -> dict:
 def datetime_from_date(
     date: Union[str, datetime.datetime, datetime.date]
 ) -> datetime.datetime:
-    """Set time to 00:00:00 in a date"""
+    """Set time to 00:00:00 in a date, keep time infor if given a datetime object"""
```
berland, line 244: typo: "infor"

tests/test_sunsch.py
```diff
@@ -725,40 +744,6 @@ def test_wrap_long_lines():
...
-    )
-
-    Path("conf.yml").write_text(
-        "files:\n  - vfp.inc\noutput: sch.inc", encoding="utf-8"
-    )
-    mocker.patch("sys.argv", ["sunsch", "conf.yml"])
-    sunsch.main()
-    schinc = Path("sch.inc").read_text(encoding="utf8")
-    assert max([len(line) for line in schinc.split("\n")]) <= 129
-    assert "Line 7 had length 146, wrapped" in caplog.text
```
berland, line 759: This is the line that fails? I would nuke only this line and keep the rest.

asnyv: yes, that is the failing line.

src/subscript/sunsch/sunsch.py
vkip: fixed

tests/test_sunsch.py
vkip: 100% correct, re-introducing the test without the last line. Verified ok with opm 2021.10.

berland:
Looks good! Suggest to rebase into two commits, one first that relaxes the  vfp_test (makes it compatible with all opm versions), then one commit that introduces clocktimes to sunsch.

vkip:
Something like this..?

berland:
Yes, assuming both commits pass all tests (I think, but have not tested). 

Ensure this is *not* squashed when it is merged (squash  and merge is default in this project), you want a merge commit since you have essentially put two PR's into one.

berland (approved):

vkip:
> Yes, assuming both commits pass all tests (I think, but have not tested).
> 
> Ensure this is _not_ squashed when it is merged (squash and merge is default in this project), you want a merge commit since you have essentially put two PR's into one.

'Merge commit not enabled for this repository'... ?

PR #474, opened 2022-03-01
Runrms modify path

jcrivenaes:
Modify paths to enable a major change in Equinor how to run rms for end users

* The old `rms` script for user will in near future be modified so that users actually run `runrms`
* This change force `runrms` to point to `/prog/roxar/rms/rms` internally
* Also, the config file for runrms shall be `/prog/res/roxapi/config/runrms.yml`. Currently this is now symlinked to `/project/res/roxari/aux/runrms.yml` for a smooth transition


ertomatic:

berland (approved):
LGTM

Issue #459, opened 2021-11-22
Stop using custom logging setup for workflows


oyvindeide:
Using: https://github.com/equinor/subscript/blob/5e159d191f393ef058e29c64f57ad9fbfd6f1156/src/subscript/__init__.py#L12 results in flooding the `stdout` of `ert` when used in workflows where `subscript` is essentially used as a library. In those cases using logging is fine, but `ert` should be responsible for setting up the logger. In those cases the standard logging module should be used directly:

```python
import logging
...
logger = logging.getLogger(__name__)
logger.info("msg")
```
for forward models using the `subscript` logger is fine, because they have their own entry points and is the module in charge of the excution.

berland:
For e.g. `csv_merge`, is this a matter of only resetting the logger object (and its level) around https://github.com/equinor/subscript/blob/5e159d191f393ef058e29c64f57ad9fbfd6f1156/src/subscript/csv_merge/csv_merge.py#L64-L72 ?

PR #482, opened 2022-05-30
Let fmuobs use the ert logger in workflow

berland:
When run as a library within ERT, fmuobs should not set its own logger, rather
it should just ask logging for a logger and let it inherit from ERTs setting

Partially solution #459


oyvindeide:
Why not split it so that if `main` is called, the `subscript` logger is used, and in `ErtScript` the standard logger is used? That will make it more visible what is going on in my opinion.

PR #490, opened 2022-08-09
Solve compatibility with Pyscal>0.9.2

berland:
* Explicitly set socr to sorw, which pyscal >0.9.2 otherwise might interpret to be larger than sorw
  This is a scenario which we don't want to support in interp_relperm
* Fix deprecation warnings from pyscal on dump_family_*()
* Remove workarounds for ancient pyscal versions (<0.7.7)


sondreso (approved):
LGTM!

PR #497, opened 2022-08-22
Handle new IndexError message for empty decks in opm 2022.04

asnyv:
Error message for requested deck keyword out of range apparently became less descriptive in `opm 2022.04`
Could alternatively keep the old message as well to support `2021.10`, but find it ok to force` 2022.04` to be able to catch it if `opm` switches back to the more descriptive message (have a feeling that it might not have been intended from `opm`'s)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -65,7 +65,7 @@ REQUIREMENTS = [
     "fmu-tools",
     "matplotlib",
     "numpy",
-    "opm==2021.10",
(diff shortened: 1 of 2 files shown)


ertomatic:

berland:
If not hard, it is beneficial to be compatible with both versions, allowing more flexibility in komodo deploy.

berland (approved):

PR #501, opened 2022-10-07
Update eclmanual default to 2022.2

asnyv:
Update default to Eclipse 2022.2


ertomatic:

asnyv:
@hhgs ref post on Yammer of updated Eclipse default. Ideally we should probably rewrite this small script to automatically fetch latest release to avoid keeping it up to date manually.

Issue #502, opened 2022-10-16
Implement script for calculating CO2 volume inside/outside a polygon boundary


vegardkv:
This addresses a specific script that already exists for post-processing simulation results from the Pflotran simulator.

Input is a result from a simulation run, which follow the same format as Eclipse (EGRID, UNRST and INIT files), as well as a polygon file (csv).

Output is a table containing the amount of CO2 (kg) inside/outside the provided polygon, with one row per time step.

If subscript is not the appropriate repo for this script, it would be nice to have discussion if there are other repos that are better suited.

mferrera:
Merged!

PR #503, opened 2022-10-16
Script for CO2 containment

vegardkv:
Aimed at #502 

As mentioned in issue description, if this repo is not suited, it would be nice to discuss if there are other existing repos that are a better fit.

Dependencies are xtgeo, numpy, pandas and shapely. Pygeos is an optional dependency that will be used for speed-up if it is available.


ertomatic:

vegardkv:
Thanks for the reply 👍 and I am aware of the issue, as well.

There are some internal discussions going on as far as I understand, either finding someone responsible for maintenance of subscript, or perhaps some other solution. I guess that means that this PR will be on hold until a conclusion has been made. Hopefully, in not too long :)

mferrera:
Closing for now as it appears that this will be superseded by a forthcoming PR from a different fork. Please inform us if this is not actually the case.

PR #507, opened 2022-12-07
Output dates in ISO-format from fmuobs

berland:
Non-ISO dates are now deprecated in ERT, so fmuobs should write to ISO-8601.


rnyb (commented):
Should the `parsers.py` file also be updated with the new variable `ERT_ISO_DATE_FORMAT`, or is that not needed?

berland:
Yes, it seems the parser also needs a corresponding update to be able to parse config files with ISO-dates. That can be possibly be done in a separate PR.

asnyv:
merge @berland @rnyb @alifbe ?

PR #516, opened 2023-01-29
Temporarily disabling ResInsight tests 

vkip:
(actually: avoiding errors by expecting them to fail) until Python interface works with Python 3.8/3.9 on latest Ubuntu packages.


ertomatic:

PR #518, opened 2023-01-31
Casegen_UpCaRs: Add option to shift origin position to anywhere in the model

alifbe:
One of the user in Equinor requested to have possibility to put the origin location in the middle of the model instead of corner. The PR will add option to put origin location at any location in the model by set the value of origin_x_pos and origin_y_pos to a value between 0 - 1 (fraction of model size in X and Y direction). In addition, user can also set the top of origin.


ertomatic:

vkip (approved):
LGTM!

PR #519, opened 2023-02-13
Let ERT pick location of stderr and stdout

berland:


berland:
This is a follow up to https://github.com/equinor/ert/pull/4671

alifbe (approved):
Looks good to me

PR #527, opened 2023-04-12
Fix compatibility issues with pandas 2.0

### LINKED ISSUE #525: test_csv2ofmvol incompatible with Pandas 2.0.0
There is a test failure when `pandas==2.0.0`:

```
$ pytest -sxv tests/test_csv2ofmvol.py::test_df2vol[dframe6-expected_lines6]
[snip]
        backagain_df = ofmvol2csv.process_volstr(volstr)
        if dframe.empty:
            assert backagain_df.empty
        else:
            # (bogus columns in dframe must be ignored)
>           pd.testing.assert_frame_equal(
                dframe[backagain_df.columns].fillna(value=0.0), backagain_df
            )

tests/test_csv2ofmvol.py:288: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pandas/_libs/testing.pyx:52: in pandas._libs.testing.assert_almost_equal
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   AssertionError: MultiIndex level [1] are different
E   
E   MultiIndex level [1] values are different (50.0 %)
E   [left]:  DatetimeIndex(['2010-01-01', '2010-01-02'], dtype='datetime64[ns]', name='DATE', freq=None)
E   [right]: DatetimeIndex(['2010-01-01', '2010-02-01'], dtype='datetime64[ns]', name='DATE', freq=None)
E   At positional index 1, first diff: 2010-01-02T00:00:00.000000000 != 2010-02-01T00:00:00.000000000

pandas/_libs/testing.pyx:172: AssertionError
================================================================================== short test summary info ===================================================================================
FAILED tests/test_csv2ofmvol.py::test_df2vol[dframe6-expected_lines6] - AssertionError: MultiIndex level [1] are different

```

### LINKED ISSUE #526: test_compare_volumetrics not compatible with Pandas 2
`$ pytest -v tests/test_rmsecl_volumetrics.py 
`
gives a lot of errors:
`TypeError: Passing a set as an indexer is not supported. Use a list instead.`

alifbe:
This PR should resolve #525 and resolve #526

### DIFF
--- a/src/subscript/ofmvol2csv/ofmvol2csv.py
+++ b/src/subscript/ofmvol2csv/ofmvol2csv.py
@@ -283,10 +283,7 @@ def parse_ofmtable(
             on_bad_lines="skip",  # pylint: disable=unexpected-keyword-arg
         )
 
-    if version.parse(pd.__version__) >= version.parse("2.0"):
-        data["DATE"] = pd.to_datetime(data["DATE"], format="mixed", dayfirst=True)
-    else:
-        data["DATE"] = pd.to_datetime(data["DATE"], dayfirst=True)
+    data["DATE"] = pd.to_datetime(data["DATE"], dayfirst=True)
 
     if "WELL" in data and "DATE" in data:
         data = data.set_index(["WELL", "DATE"]).sort_index()
--- a/src/subscript/ofmvol2csv/ofmvol2csv.py
+++ b/src/subscript/ofmvol2csv/ofmvol2csv.py
@@ -283,7 +283,10 @@ def parse_ofmtable(
             on_bad_lines="skip",  # pylint: disable=unexpected-keyword-arg
         )
 
-    data["DATE"] = pd.to_datetime(data["DATE"], dayfirst=True)
+    if version.parse(pd.__version__) >= version.parse("2.0"):
+        data["DATE"] = pd.to_datetime(data["DATE"], format="mixed", dayfirst=True)
+    else:
+        data["DATE"] = pd.to_datetime(data["DATE"], dayfirst=True)
 
     if "WELL" in data and "DATE" in data:
         data = data.set_index(["WELL", "DATE"]).sort_index()
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ __pycache__/
 
 # Distribution / packaging
 .Python
-env/
+env*/
 build/
 develop-eggs/
 dist/
--- a/src/subscript/rmsecl_volumetrics/rmsecl_volumetrics.py
+++ b/src/subscript/rmsecl_volumetrics/rmsecl_volumetrics.py
@@ -76,7 +76,7 @@ def _compare_volumetrics(
     absolute differences of eclipse volumes minus RMS volumes.
     """
     set_data_list = []
-    for set_idx, regzonfip_set in disjoint_sets_df.groupby(["SET"]):
+    for set_idx, regzonfip_set in disjoint_sets_df.groupby("SET"):
         set_results = {"SET": set_idx}
         # Slice and sum simvolumes_df for the FIPNUMS in this set:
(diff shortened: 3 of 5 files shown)


ertomatic:

src/subscript/ofmvol2csv/ofmvol2csv.py
```diff
@@ -283,7 +283,10 @@ def parse_ofmtable(
             on_bad_lines="skip",  # pylint: disable=unexpected-keyword-arg
         )
 
-    data["DATE"] = pd.to_datetime(data["DATE"], dayfirst=True)
+    if version.parse(pd.__version__) >= version.parse("2.0"):
+        data["DATE"] = pd.to_datetime(data["DATE"], format="mixed", dayfirst=True)
+    else:
+        data["DATE"] = pd.to_datetime(data["DATE"], dayfirst=True)
```
berland, line 289: Can you confirm if the test that requires this branching is actually relevant?

alifbe: Unfortunately yes. pandas 1.5.3 doesn't accept format="mixed"  and will fail hard.

alifbe: @berland or do you mean we can prune some of the tests that failed without this branching? @rnyb is reaching out to OFM responsible to check if OFM has a specific date format.

berland: That is what I am asking, check whether the test code is doing meaningless tests. Parsing "2022-12-24" with dayfirst=True trying to get Pandas to think that this is day 12 and month 24 would be an example of such a meaningless test.

alifbe: Removing this branching will result in 3 failures. 

Assuming that OFM will not generate a mixed date format within a file, two of the tests can probably be removed:
https://github.com/equinor/subscript/blob/87c74b52819d90057bd9690c7f2db75152840759/tests/test_ofmvol2csv.py#L238-L251
https://github.com/equinor/subscript/blob/87c74b52819d90057bd9690c7f2db75152840759/tests/test_ofmvol2csv.py#L326-L340

But this branching also solve the issue that I mentioned in #525 

https://github.com/equinor/subscript/blob/87c74b52819d90057bd9690c7f2db75152840759/tests/test_csv2ofmvol.py#L224-L240

berland: Agree that the two first tests are probably not relevant, and is not a good enough reason to have the pandas-version-branch in the code.

alifbe: I just make a few changes:
- Revert pandas version branching
- Remove first test above
- Modify second test above to consistent date format since the test has other purpose (Well encoded as a column)
- Update the test so that it also use dayFirst=True, which make the 3rd test comparable.

berland (approved):
Looks good! Ensure it is squashed when merged, and ensure the commit message is updated accordingly from the two existing commit messages.

PR #528, opened 2023-04-21
Sector2fluxnum optional eclipse version

### LINKED ISSUE #512: [sector2fluxnum] Default eclipse version should be bumped
sector2fluxnum uses Eclipse version 2014.2 by default in tests:
https://github.com/equinor/subscript/blob/master/tests/test_sector2fluxnum.py#L39

and is also a default in the code
https://github.com/equinor/subscript/blob/b6c2e076946487a7d8f663ab3b1f75f8c41fc446/src/subscript/sector2fluxnum/datafile_obj.py#L393

This Eclipse version is not available in the Azure compute nodes, and must be bumped.

alifbe:
Make sector2fluxnum ecl_version argument optional. If ecl_version is not specified, runeclipse will run a using a default version (currently 2021.3)

Resolve #512

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,7 +8,7 @@ __pycache__/
 
 # Distribution / packaging
 .Python
-env/
+env*/
 build/
 develop-eggs/
 dist/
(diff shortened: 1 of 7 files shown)


ertomatic:

PR #531, opened 2023-04-24
Pin numpy version

alifbe:
Temporarily pin numpy to 1.24.2


ertomatic:

PR #536, opened 2023-04-28
Auto update version for `eclmanual` and `runeclipse`

alifbe:
Automatic fetch latest Eclipse version and using for `eclmanual` and `runeclipse`

The PR fetch E100 version and assumes that there is a corresponding E300 and Eclipse manual.


ertomatic:

PR #539, opened 2023-05-11
Update location of testkomodo.sh

berland:
The location inside jenkins has been deprecated in komodo-releases


alifbe:
@berland looks like test is broken due to new `opm` (v2023.04) library. Do you know which repo I should look at to see the changes? The last successful test was using opm 2022.10 

I look at https://github.com/OPM/opm-common/releases but their latest release is 2022.10

berland:
To have the build pipeline running, you can pin opm to a previous releasev @alifbe

mferrera:
The OPM package is updated and works correctly now (but a few tests fail against its new behavior)

PR #543, opened 2023-05-30
Activate nightly test

alifbe:
Activate nightly test again

### DIFF
--- a/.github/workflows/subscript.yml
+++ b/.github/workflows/subscript.yml
@@ -10,17 +10,19 @@ on:
   release:
     types:
       - published
+  schedule:
+    # Run CI every night and check that tests are working with latest dependencies
+    - cron: "0 0 * * *"
 
 env:
   MPLBACKEND: Agg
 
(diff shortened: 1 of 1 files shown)


.github/workflows/subscript.yml
```diff
@@ -81,22 +83,22 @@ jobs:
       - name: Update GitHub pages
         if: github.repository_owner == 'equinor' && github.ref == 'refs/heads/master' && matrix.python-version == '3.6'
```
rnyb, line 84: `matrix.python-version == '3.6'`
Should this be updated?

rnyb (approved):
Your updates looks good, but added a question to another part of the file that may need to be adressed?

.github/workflows/subscript.yml
alifbe: That's true....Good catch. I checked the history, we used to test subscript against Python 3.6 as well. I have now updated it as well.

PR #545, opened 2023-05-31
Give more helpful encoding exceptions in pack_sim

### LINKED ISSUE #537: pack_sim issues with none UTF files
if any Norwegian letters is present in the DATA file ( like a comment on any lines explaining anything in Norwegian), the code exits with an error message.

Solutions:
1. to make the code with any file type
2. have a temporary copy of the DATA file in UTF and then delete it afterwards.

### LINKED ISSUE #538: pack_sim issue with comments after main commands
As obvious, Eclipse DATA file is separated into sections by commands like "RUNSPEC, GRID, EDIT, PROPS, REGIONS, SOLUTION, SUMMARY and SCHEDULE.

when -fmu option is selected and pack_sim is supposed to put include files after each section into a folder by the same section name,

if any comments is entered after these commands, it will not be recognized. for example:

GRID  -- comment

is not recognized and the files will be added to the previous section if that one is recognized. if all the main sections have comment, then the result is not very understandable.

mferrera:
Resolves #537 

- Added pytest-xdist to parallelize the tests
- Pinned urllib3<2, as urllib3>=2 is pulled in through the dependency tree and is incompatible with RGS machines
- Gave a more helpful exception reason on utf-8 encoding errors. It roughly copies how ERT handles it but stops at the first bad character

A specific command to do the conversion _could_ be given but it's probably better not to be responsible if it damages the file.

### DIFF
--- a/.github/workflows/subscript.yml
+++ b/.github/workflows/subscript.yml
@@ -67,7 +67,7 @@ jobs:
 
       - name: Run tests
         run: |
-          pytest tests
+          pytest -n auto tests
           # Check that repository is untainted by test code:
           git status --porcelain
           test -z "$(git status --porcelain)"
--- a/README.rst
+++ b/README.rst
@@ -48,7 +48,7 @@ In a fresh virtual environment you should be able to do::
 
 and all dependencies should be installed. Confirm your installation with::
 
-  pytest
+  pytest -n auto
 
 and this should run for some minutes without failures.
 
--- a/docs/contribution.rst
+++ b/docs/contribution.rst
@@ -61,7 +61,7 @@ repository, which you can do by running:
 
 .. code-block:: console
 
-  pytest
+  pytest -n auto
 
 
 Repository conventions
--- a/setup.py
+++ b/setup.py
@@ -75,6 +75,7 @@ REQUIREMENTS = [
     "scipy",
     "seaborn",
     "segyio",
+    "urllib3<2",
     "xlrd",
     "xtgeo",
 ]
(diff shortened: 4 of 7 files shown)


alifbe (commented):
@mferrera very nice 👍 

Do we need to check it as well for `inspect_file` and `_md5checksum` as well?

Also would it be nice/needed to print out the correct encoding to end-user? I see that komodo 
 has `chardet` library install which can guess the correct encoding format or alternatively `file -ib`

mferrera:
I don't think we need to check in `inspect_file` or `_md5checksum`. The root `.DATA` file won't make it to `inspect_file` with an invalid character, but it is possible that `INCLUDE`d files have invalid characters. The script already checks for this but concludes that they are actually binary and just copies them without extra processing, so it won't fail (but maybe it should). I added another test case for this.

There are a number of possible improvements that could address possibly-unintentional logic like this in the script but maybe out of scope.

I'm not certain most users would gain a lot of actionable insight if we tell them the encoding. My guess is that the vast majority will probably find ISO-8859-1 a bit meaningless to them 😅 but if you think it's a good idea I'm happy to add it

alifbe (approved):
@mferrera looks good to me.

Regarding printing out correct encoding, I’m not sure either. I saw in slack channel that user can change/convert the file encoding by using iconv command and specify which encoding to convert from.

alifbe (approved):
@mferrera looks good to me.

Regarding printing out correct encoding, I’m not sure either. I saw in slack channel that user can change/convert the file encoding by using iconv command and specify which encoding to convert from.

PR #548, opened 2023-06-05
pack_sim - Properly check for binary files

### LINKED ISSUE #547: [pack_sim] Give a better binary file heuristic
Currently pack_sim concludes a file is binary if it fails to be utf-8 decoded, which means it will copy over iso-8859-1 files without inspection. This means that it could potentially skip over other included files and generally cause the package to be incomplete.

One way this could be handled is by reading an arbitrary number of bytes, e.g. 512, and looking for NUL-bytes that are unlikely to be from UTF-16 encoding.

mferrera:
Resolves #547 

Use the binary file check heuristic from `eclcompress` and read lines from a wrapper function that ensures a file is utf-8. This also ensures that recursively-many file handles aren't held during execution.

### DIFF
--- a/src/subscript/pack_sim/pack_sim.py
+++ b/src/subscript/pack_sim/pack_sim.py
@@ -10,6 +10,7 @@ from shutil import copy
 from typing import Dict, List, Optional, TextIO, Union
 
 from subscript import __version__, getLogger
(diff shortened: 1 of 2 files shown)


rnyb (approved):
LGTM

PR #549, opened 2023-06-16
Pin Sphinx < 7 and build without setuptools hook

mferrera:
Sphinx >= 7 deprecates building docs through its setuptools hook and deprecates the `style` HTML key that is used by the sphinx_rtd_theme. The theme has not yet been updated to address this change.

### DIFF
--- a/.github/workflows/subscript.yml
+++ b/.github/workflows/subscript.yml
@@ -26,15 +26,12 @@ jobs:
 
     steps:
       - name: Checkout commit locally
-        uses: actions/checkout@v2
-
-      - name: Checkout tags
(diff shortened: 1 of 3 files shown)


alifbe:
@mferrera will this resolve the issue that Sondre flag?

mferrera:
Yep!

mferrera:
It seems most other repositories have already pinned Sphinx. I have seen it pinned to 4, 5, 6 in various repositories. I'm more surprised that subscript has lasted so long without pinning 😅

alifbe (approved):
Looks good to me and good job with updating action version. 👍

PR #551, opened 2023-06-16
Change master to main

mferrera:


alifbe (approved):
LGTM 👍 

I did a search for `master` in `subscript` repo and found this: https://github.com/equinor/subscript/blob/e5e3f72a4600d86c6bdf7f79787f6595da792514/.pre-commit-config.yaml#L6 Not sure if this also need to be replaced.

mferrera:
Nice catch. I fixed it, although the branch protection rules should prevent it from getting upstream anyway.

PR #553, opened 2023-06-20
Legacy nosim - insert NOSIM in RUNSPEC

### LINKED ISSUE #424: [nosim] does not work if TITLE is missing in deck
~~If a deck contains~~
```
-- NOSIM
```

~~then nosim will not modify the deck, but still continue to run the deck in interactive mode, but then not in "nosim"-mode.~~

mferrera:
Resolves #424 

This copies the regex used in semeio.


rnyb (approved):
lgtm

PR #554, opened 2023-06-21
fmu_copy_revision - Add empty directory test

### LINKED ISSUE #428: [fmu_copy_revision] skips empty folders
The new fmu_copy_revision (python) will not copy over empty directories, whereas the perl version does.

This has not yet been reported by users, but might be a problem some day?

mferrera:
Resolves #428 

It seems that fmu_copy_revision runs as expected for the provided profile type. Users may not have realized the profile they selected does not copy empty directories so it may become more "fixed" with added documentation.

Regarding the comment about the contents of /output directories, it seems to have been fixed in [85ee037](https://github.com/equinor/subscript/commit/85ee037554271374031c31f8526c1e0eb6a8d0ba)

### DIFF
--- a/src/subscript/fmu_copy_revision/fmu_copy_revision.py
+++ b/src/subscript/fmu_copy_revision/fmu_copy_revision.py
@@ -640,20 +640,7 @@ class CopyFMU:
         ]
         logger.info(" ".join(command))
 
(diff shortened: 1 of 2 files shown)


alifbe (approved):
Looks good to me 👍

PR #555, opened 2023-06-21
Update check_swatinit due to Flow support PPCWMAX

alifbe:
Latest OPM Flow (June 16 2023) supports PCCWMAX keyword. But it gives a slightly different results compare to Eclipse due to how fine the calculation is. OPM Flow seems to be more accurate in the calculation.

However, since official release flow is still 2023.04, the test will fail. Hence disable it until OPM release a new version.

### DIFF
--- a/tests/test_check_swatinit_simulators.py
+++ b/tests/test_check_swatinit_simulators.py
@@ -117,48 +117,27 @@ def test_swat_limited_by_ppcwmax_above_contact(simulator, tmp_path):
     qc_frame = run_reservoir_simulator(simulator, model)
     assert np.isclose(qc_frame["PPCWMAX"][0], 3.01)
     qc_vols = qc_volumes(qc_frame)
-    swat_if_ppcwmax = 0.5746147
     if "eclipse" in simulator:
         # for PPCWMAX set to 3.01 Eclipse100 will scale the swatinit value to this:
-        assert qc_frame["QC_FLAG"][0] == __PPCWMAX__
-        assert np.isclose(qc_frame["SWAT"][0], swat_if_ppcwmax)
-        assert np.isclose(qc_frame["PC_SCALING"][0], 3.01 / 3.00)
         actual_pc = 1.4226775
-        assert np.isclose(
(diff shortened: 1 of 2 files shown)


tests/test_check_swatinit_simulators.py
```diff
@@ -117,48 +120,28 @@ def test_swat_limited_by_ppcwmax_above_contact(simulator, tmp_path):
...
-            # Worryingly inaccurate?
-            model.evaluate_sw(actual_pc, scaling=3.01 / 3.00),
-            swat_if_ppcwmax,
-            atol=0.02,
-            # (opm probably uses monotone cubic interpolation in SWOF, but that
-            # should not affect SWOF tables with only two points)
-        )
+        # OPM Flow gives different swat due to different resolution
+        # in number of sections to calculate equilibrium
+        # Ref: email from Torbjørn Skille
```
berland, line 130: The named reference is of limited use in the source code repository, only the explanation itself is verifiable (it sounds a little bit like FINE_EQUIL being enabled by default?)

alifbe:
> Recommend to squash the last fixup commit into the first one, and keep the second commit. The second commit should be kept so that it becomes easy to revert later

Good point. 👍

tests/test_check_swatinit_simulators.py
alifbe: I'm not sure what is FINE_EQUIL that you are referring.

The explanation from Torbjørn is that:

> Capillary pressure are calculated from densities which are dependent on pressure. The pressure will vary with depth and this is of course not linear. I have not programmed this myself but the depth interval covered with the grid needs to be divided into  a finite number of sample points and capillary pressures are then calculated for these points based on pressure and capillary pressure specified at datum depth.  The number of sampling points is not exposed for users, nightery in Eclipse of OPM flow. This is happening inside the simulator. 

I think he is referring to this code:
https://github.com/OPM/opm-simulators/blob/815bb7a49369a1a765cc5f9836144a8833c5d5aa/ebos/equil/initstateequil.hh#L169 

When he varied these number of points, he found that the Pc from Eclipse is obtained when using 400 sections and OPM is using 2000 sections by default
image

PR #561, opened 2023-06-22
Move from setup.py to pyproject.toml

### LINKED ISSUE #557: Add pyproject.toml and rework top-level dot/config files
pyproject.toml is required to upload to PyPI and while adding it we could clean up a few other things in the top level.

- Remove `.libecl_version` -- appears to be a legacy file from travis CI days
- Split `setup.py` into `pyproject.toml`, as much as possible
- See if mypy/pylint configurations are better merged into pyproject
- Update pre commit hooks
- Double-check MANIFEST.in

mferrera:
Resolves #557 

pyproject.toml is the home of the more modern and recommended build system and required to be uploaded to PyPI.

Some files were deleted as moved into there, and pre commit was deleted because I doubt we are using and it and CI catches it. But correct me if I'm wrong!


berland:
The file ci/testkomodo.sh also requires a slight update in this PR.

berland:
I am unsure of the importance of MANIFEST.in, it does not look like its instructions are in the new pyproject.toml?

mferrera:
From what I gathered MANIFEST.in is not necessary when setuptools_scm is being used. It helps to include files that may not be included in a distribution by default (i.e., not Python modules/packages) [[1]](https://packaging.python.org/en/latest/guides/using-manifest-in/), but setuptools_scm supersedes this by using building distributions that are tracked by git. [[2]](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html) (third paragraph)

alifbe (approved):
👍

PR #567, opened 2023-06-27
Ensure /global/bin is in PATH for komodo

berland:
The test environment in which pytest is run when doing komodo tests is limited. This commit will ensure that /global/bin in in the PATH. The added instruction is idempotent.

/global/bin is needed as it is the typical location for eclrun which runeclipse in subscript now uses to locate the needed binaries for Eclipse.


alifbe (approved):
LGTM 👍

PR #568, opened 2023-06-27
 Script for CO2 containment

AudunSektnanNR:
Script for calculating the amount of CO2 inside and outside a given perimeter, and separates the result per formation and phase (gas/dissolved). Output is a table on CSV format.

Can take both a containment polygon and a "hazardous" polygon. Result is divided into the regions inside vs outside vs hazardous, where inside means inside the containment polygon.

Can calculate either CO2 mass or a representative CO2 volume (three different calculation methods for volume). CO2 mass is the most used and tested.

The resulting CSV-file can be used as data in the CO2Leakage webviz plugin.

See (old) issue: #502


src/subscript/co2containment/calculate.py
```diff
@@ -0,0 +1,218 @@
...
+        x_coord (np.ndarray): x coordinates
+        y_coord (np.ndarray): y coordinates
+        poly (Union[Polygon, MultiPolygon]): The polygon that determines the
+                                             containment of the (x,y) coordinates
+
+    Returns:
+        np.ndarray
+    """
+    return np.array([poly.contains(Point(_x, _y)) for _x, _y in zip(x_coord, y_coord)])
+    # try:
```
alifbe, line 209: Are these redundant?

src/subscript/co2containment/co2_calculation.py
```diff
@@ -0,0 +1,820 @@
...
+                                     volume_actual/volume_actual_simple)
+        co2_molar_mass (float): CO2 molar mass - Default is 44 g/mol
+        water_molar_mass (float): Water molar mass - Default is 18 g/mol
+
+    Returns:
+      Co2Data
+    """
+    # pylint: disable-msg=too-many-locals
+    # pylint: disable-msg=too-many-branches
+    # pylint: disable-msg=too-many-statements
```
alifbe, line 620: Just a suggestion, you could consider refactor the code to have less branches so that it's easier to read.

alifbe (changes requested):
Please fix the order of import (using `isort `or `vscode`)

src/subscript/co2containment/calculate.py
AudunSektnanNR: The code using Pygeos is an alternative code with speed-up, if Pygeos is available. I can remove this code for now.

src/subscript/co2containment/co2_calculation.py
AudunSektnanNR: Probably a good idea, maybe do that sometime in the future.

docs/scripts/co2_containment.rst
```diff
@@ -0,0 +1,48 @@
+
+CO2_CONTAINMENT
+==============
+
+Calculates the amount of CO\ :sub:`2` inside and outside a given perimeter, and separates the result per formation and phase (gas/dissolved). Output is a table on CSV format.
+
+Input
```
mferrera, line 7: Please consider using 

```rst
.. argparse::
   :module: subscript.co2_containment.co2_contaiment
   :func: get_parser
   :prog: co2_containment
```

Sphinx can create a nice listing of the inputs from the options supplied to argparse -- see the other scripts which do this. It helps to keep documentation up-to-date by minimizing the number of places it needs to be updated. But whichever may give the most information

AudunSektnanNR:
Have fixed some of the PR issues. Still some error from mypy-checks. Looking into that.

src/subscript/co2containment/calculate.py
```diff
@@ -0,0 +1,218 @@
+"""CO2 calculation methods"""
```
mferrera, line 1: Please rename the directory to `subscript/co2_containment` match the entry script filename and add an empty `__init__.py` file as the other scripts have.

subscript has the (intentional or not, perhaps dubious) convention of matching package/module names `subscript.script_a.script_a`

src/subscript/co2containment/co2_containment.py
```diff
@@ -0,0 +1,346 @@
...
+        df4e = df4.loc["gas", "hazardous"].rename(columns={"amount": "gas_hazardous"})
+        df4f = df4.loc["aqueous", "hazardous"].rename(
+            columns={"amount": "aqueous_hazardous"}
+        )
+        for _df in [df2a, df2b, df3a, df3b, df3c, df4a, df4b, df4c, df4d, df4e, df4f]:
+            total_df = total_df.merge(_df, on="date", how="left")
+    return total_df.reset_index()
+
+
+def make_parser() -> argparse.ArgumentParser:
```
mferrera, line 204: Not super important, but consider renaming to `get_parser()` as this is what the vast majority of other subscripts call it. Doesn't matter so much so long as the args receiver in the documentation (if used) has the right name.

tests/test_co2_calculate.py
```diff
@@ -0,0 +1,256 @@
...
+def _xy_and_volume(grid: xtgeo.Grid):
+    """
+    Get xy and volume
+    """
+    xyz = grid.get_xyz()
+    vol = grid.get_bulk_volume().values1d.compressed()
+    return xyz[0].values1d.compressed(), xyz[1].values1d.compressed(), vol
+
+
+# @pytest.fixture
```
mferrera, line 45: Leftover comment?

tests/test_co2_containment.py
```diff
@@ -0,0 +1,103 @@
...
+
+from subscript.co2containment.co2_calculation import (
+    SourceData,
+    _calculate_co2_data_from_source_data,
+    CalculationType,
+    Co2Data,
+)
+
+
+# @pytest.fixture
```
mferrera, line 13: Leftover comment?

tests/test_co2_containment.py
```diff
@@ -0,0 +1,103 @@
...
+        DGAS={date: np.ones(size) * 100.0 for date in dates},
+        AMFG={
+            date: np.ones(size) * 0.02 * value
+            for date, value in gas_saturations.items()
+        },
+        YMFG={date: np.ones(size) * 0.99 for date in dates},
+    )
+
+
+# @pytest.fixture
```
mferrera, line 51: Not to be pedantic with these, but another one :)

mferrera (changes requested):

mferrera:
Looking good! 😄 Thanks a lot for adapting to the linters, a bit of a gauntlet to be sure.

My last concern is with the test coverage -- it becomes our third outlier at significantly below 80% coverage and more importantly we lack the CO2 domain knowledge that may prevent us from addressing issues ourselves. I guess normally coverage is the sort of thing that increases over time as users do unexpected user things, but because we are doing a change of hands it is a bit more of a worry.

Being aware that you have a time runway to work with, do you think there are any critical or highly domain specific methods that could potentially add to long-term robustness?

alifbe (approved):
LGTM 👍

PR #570, opened 2023-06-29
eclcompress - compress only allowlist keywords by default

### LINKED ISSUE #499: ECLCOMPRESS damages PVT file (PVTO table)
ECLCOMPRESS acts on Eclipse include file with PVTO table, and some other Eclipse include files in eclipse/include/props folder. These files are not in grdecl format. Get an error in Eclipse due to damaged (compressed) PVTO table.

Also, the documentation on ECLCOMRESS in Subscript is not consistent. Should update text for the "file" argument.

### LINKED ISSUE #566: Enhance ECLCOMPRESS documentation included in internal ERT docs
https://github.com/equinor/subscript/blob/7e1de2a9e1669df2d253232da2ebd62012918f1e/src/subscript/eclcompress/eclcompress.py#L19C1-L25C4

Add a note that ```--files/<FILES>``` expects to receive a file containing a list of file paths to glob from.

mferrera:
Resolves #566 
Resolves #499

The allowlist was generated by parsing the Eclipse reference manual for keywords that allow repeated counts. It can now compress multiple keywords within a multi-keyword context (e.g. `EQUALS`) as well as keywords whose contents and terminator are on the same line.

It defaults to the allowlist but still allows a regex to be applied. This removes the old way of functioning in which keywords were evaluated only by a regex which could in some cases apply compression to keywords which do not allow repeat counts and thereby damage the file.


src/subscript/eclcompress/eclcompress.py
```diff
@@ -312,19 +339,27 @@ def find_keyword_sets(
...
-            kwstart = lineidx
+        line = line.strip()
+        if not line:
             continue
-        if kwstart is not None and line.strip()[0:2] == "--":
+        if eclkw_regexp:
+            if re.match(eclkw_regexp, line) and line not in DENYLIST_KEYWORDS:
+                kwstart = lineidx
+                continue
+        else:
```
alifbe, line 353: @mferrera does it mean that if user specify the regular expression, it will ignore the ALLOWLIST_KEYWORDS ? 

I would rather this to be in addition because the ALLOWLIST_KEYWORDS are like safelist. I think this behavior is more relevant earlier because there is no safe list so this argument will be a way to define safelist. What do you think?

alifbe (approved):
Great job. 👍 

Please consider my comment regarding regular expression on user command line argument.

src/subscript/eclcompress/eclcompress.py
mferrera: It does and I agree that's probably the most sane thing to do. My thought was to preserve the old behavior if a regex is provided in case someone is using it in unexpected ways, i.e. they only want to compress one particular keyword or something. I think it keeps the script consistent with the old behavior while ensuring the default behavior doesn't cause harm.

I'm actually skeptical this regex feature is used at all and maybe we should consider deprecating it, especially with having a complete list of compressible keywords now. At the very least it is not a usable option when running from ERT.

berland: I would also be surprised if anyone is actually using the regex feature.

PR #577, opened 2023-07-06
Let argparse handle co2_containment args

mferrera:
When invoked from the command line via its script name co2_containment would fail to start as `main()` expected arguments to be passed into it from a direct interpreter invocation. This change lets argparse handle invocation arguments without passing them to it directly.


alifbe (approved):
Looks good to me and more consistent with other scripts. 👍

PR #578, opened 2023-07-18
Create test files in temporary directories

mferrera:
A few tests left test files in the root directory after a test run. This could cause race conditions in some cases with pytest-xdist and tests should not leave junk files in the root directory at any rate.


alifbe:
@mferrera Nice.....I did a quick search and found a few other places as well which are not typically run in the local setup, but I think is run within `komodo` testing. Do you think it's worth to update as well?

https://github.com/equinor/subscript/blob/e4cba62e44aaf50b0c0e5ea9906f01bf60a58a72/tests/test_ofmvol2csv.py#L513
https://github.com/equinor/subscript/blob/e4cba62e44aaf50b0c0e5ea9906f01bf60a58a72/tests/test_check_swatinit_simulators.py#L57

mferrera:
Good catch! The ones you highlighted are actually okay -- they use a fixture that changes dir for them (former case) or call the simulator function after changing dir in the test function (latter case). But I found a few cases where the dir change didn't happen or otherwise left files around. I don't seem to find any more files in the root dir when run with simulators as well now.

alifbe:
> Good catch! The ones you highlighted are actually okay -- they use a fixture that changes dir for them (former case) or call the simulator function after changing dir in the test function (latter case). But I found a few cases where the dir change didn't happen or otherwise left files around. I don't seem to find any more files in the root dir when run with simulators as well now.

Good to know 😄

alifbe (approved):
LGTM 👍

PR #579, opened 2023-07-21
Add Eclipse keyword regex deprecation warning to eclcompress

### LINKED ISSUE #575: Deprecate user provided regex in `eclcompress`
It's unlikely this functionality is used and it offers more risks than benefits now that we have an allow list of valid keywords to compress.

Given its unlikely usage I propose issuing a shorter deprecation timeline of three komodos after our next release in August. Therefore a deprecation notice will be issued if this flag is used beginning in the August komodo with removal in November komodo.

(I also propose "komodo" as a unit of time as it is sometimes in-congruent with months 😄)

mferrera:
Resolves #575

### DIFF
--- a/src/subscript/eclcompress/eclcompress.py
+++ b/src/subscript/eclcompress/eclcompress.py
@@ -9,6 +9,7 @@ import os
 import re
 import shutil
 import textwrap
+import warnings
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


alifbe (approved):
LGTM

PR #581, opened 2023-07-21
Add FORWARD MODEL and workflow job for params2csv

alifbe:
- Add glob to parameterfile argument (needed so that it works from within ERT)
- Add forward model PARAMS2CSV
- Add workflow job PARAMS2CSV

### DIFF
--- a/src/subscript/config_jobs/PARAMS2CSV
+++ b/src/subscript/config_jobs/PARAMS2CSV
@@ -3,4 +3,4 @@ EXECUTABLE  params2csv
 DEFAULT  <OUTPUT> parameters.csv
 DEFAULT  <FILENAMECOLUMN> filename
 
-ARGLIST  "--filenamecolumnname" <FILENAMECOLUMN> "-o" <OUTPUT> <PARAMETERFILES> "--keepconstantcolumns"
+ARGLIST  "--verbose" "--filenamecolumnname" <FILENAMECOLUMN> "-o" <OUTPUT> <PARAMETERFILES> "--keepconstantcolumns"
--- a/src/subscript/config_jobs/PARAMS2CSV
+++ b/src/subscript/config_jobs/PARAMS2CSV
@@ -1,0 +1,6 @@
+EXECUTABLE  params2csv
+
+DEFAULT  <OUTPUT> parameters.csv
+DEFAULT  <FILENAMECOLUMN> filename
+
+ARGLIST  "--filenamecolumnname" <FILENAMECOLUMN> "-o" <OUTPUT> <PARAMETERFILES> "--keepconstantcolumns"
--- a/src/subscript/params2csv/params2csv.py
+++ b/src/subscript/params2csv/params2csv.py
@@ -44,13 +44,11 @@ This forward model will convert all keys in `parameters.txt` to columns in
 In addition, it will add a column `filename` which list the source parameters.txt file. 
 This column will be useful when <PARAMETERFILES> contains wildcards.
 
-The `filename` column can be renamed by adding an argument <FILENAMECOLUMN> to the 
-FORWARD_MODEL.
+The `filename` column can be renamed by adding an argument <FILENAMECOLUMN> to the FORWARD_MODEL.
     
 .. code-block:: console
 
-  FORWARD_MODEL PARAMS2CSV(<PARAMETERFILES>=parameters.txt, <OUTPUT>=parameters.csv, 
-  <FILENAMECOLUMN>=source_file)
+  FORWARD_MODEL PARAMS2CSV(<PARAMETERFILES>=parameters.txt, <OUTPUT>=parameters.csv,<FILENAMECOLUMN>=source_file)
   
 """  # noqa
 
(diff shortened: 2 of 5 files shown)


src/subscript/params2csv/params2csv.py
```diff
@@ -68,18 +138,30 @@ def get_parser() -> argparse.ArgumentParser:
...
+
+    Args:
+        args (argparse.Namespace): Namespace with command line arguments
+    """
+    if args.verbose:
+        logger.setLevel(logging.INFO)
+
+    # Expand wildcards if not being expanded
+    args.parameterfile = [
+        path for pattern in args.parameterfile for path in sorted(glob(pattern))
```
mferrera, line 163: More curious than anything, but does sorting the globbed files cause any differences?

src/subscript/params2csv/params2csv.py
```diff
@@ -103,14 +183,12 @@ def main() -> None:
...
-                + ", not writing"
+            logger.info(
+                "Column name %s was already in %s, not writing",
+                args.filenamecolumnname,
+                parameterfilename,
             )
-            print(
+            logger.info(
                 "this filename into CSV output. Use --filenamecolumnname to avoid this."
             )
```
mferrera, line 193: Could these two log statements be collapsed to one?

mferrera (approved):
Looks great! Please consider the comments and remember to squash or rebase before merging 😄

src/subscript/params2csv/params2csv.py
```diff
@@ -5,13 +5,68 @@
...
+In addition, it will add a column `filename` which list the source parameters.txt file. 
+This column will be useful when <PARAMETERFILES> contains wildcards.
+
+The `filename` column can be renamed by adding an argument <FILENAMECOLUMN> to the 
+FORWARD_MODEL.
+    
+.. code-block:: console
+
+  FORWARD_MODEL PARAMS2CSV(<PARAMETERFILES>=parameters.txt, <OUTPUT>=parameters.csv, 
+  <FILENAMECOLUMN>=source_file)
```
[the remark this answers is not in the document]
alifbe: You are right. It should be one line :blush: Thanks

src/subscript/params2csv/params2csv.py
```diff
@@ -27,6 +82,21 @@ class CustomFormatter(
...
+    to be used as an ERT workflow (wrapping the command line utility)"""
+
+    # pylint: disable=too-few-public-methods
+    def run(self, *args):
+        # pylint: disable=no-self-use
+        """Parse with a simplified command line parser, for ERT only,
+        calling params2csv_main()"""
+        parser = get_parser()
+        args = parser.parse_args(args)
+        logger.setLevel(logging.INFO)
```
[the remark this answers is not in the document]
alifbe: I copied this behavior from `csv_merge` and thought it was a good idea to have more print out information. But now that you point it out, it does make more sense in the workflow file which unfortunately has to be created by user manually. 

I will remove this and put `--verbose` in as an example.

src/subscript/params2csv/params2csv.py
alifbe: I don't think it matter in this case. I was trying to replicate the unix glob behavior which seems to be sorted by creating time and to some level in order (probably because ERT created these parameters.txt in sequential)

Also, feels like it's nicer to see them sorted in the csv file :smile:

src/subscript/params2csv/params2csv.py
alifbe: Yes, it's definitely a good point. :+1:

src/subscript/config_jobs/PARAMS2CSV
```diff
@@ -0,0 +1,6 @@
+EXECUTABLE  params2csv
+
+DEFAULT  <OUTPUT> parameters.csv
+DEFAULT  <FILENAMECOLUMN> filename
+
+ARGLIST  "--filenamecolumnname" <FILENAMECOLUMN> "-o" <OUTPUT> <PARAMETERFILES> "--keepconstantcolumns"
```
mferrera, line 6: It occurred to me as well that maybe `--verbose` should go in here by default as well? What do you think?

src/subscript/params2csv/params2csv.py
```diff
@@ -103,16 +180,13 @@ def main() -> None:
...
-            )
-            print(
-                "this filename into CSV output. Use --filenamecolumnname to avoid this."
+            logger.info(
+                "Column name %s was already in %s, not writing this filename "
+                "into CSV output. Use --filenamecolumnname to avoid this.",
+                args.filenamecolumnname,
+                parameterfilename,
             )
+            logger.info()
```
mferrera, line 189: This empty one could probably be removed entirely now as well 😄

mferrera (approved):
Thanks for considering the comments! Just some quick follow-ups

src/subscript/config_jobs/PARAMS2CSV
alifbe: Totally support that.

src/subscript/params2csv/params2csv.py
alifbe: of course :rofl:

PR #582, opened 2023-08-10
Update workflows and add PyPI publish workflow

### LINKED ISSUE #249: Deploy to pypi
Use `equinor-subscript` as name on PyPI, we  should not grab the too-general `subscript` on pypi. 

Consider implications on the python package name, and with Komodo.

### LINKED ISSUE #576: Run subscript against Python 3.10, 3.11 in CI
We should anticipate a crawl toward higher Python versions in the future

mferrera:
subscript is now available at https://pypi.org/project/subscript/. The new workflow uses PyPI's trusted publisher mechanism which checks for the presence of `publish.yml` in this repository in this organization and does some magic to accept distributions from it without a token.

As a consequence this workflow _cannot_ be renamed without adapting the PyPI project. It is set to trigger only when a release tag is pushed -- but we should test it.

We should add another maintainer to the PyPI project for some redundancy :-)

Resolves #249
Resolves #576

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,33 @@
+name: Publish to PyPI
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  pypi-publish:
+    name: Upload release to PyPI
+    runs-on: ubuntu-latest
+    environment:
+      name: pypi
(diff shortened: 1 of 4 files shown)


alifbe (approved):
Looks good to me 👍

Issue #583, opened 2023-08-10
`pack_sim` wrongly adds an extra INCLUDE


rnyb:
When running `pack_sim` an extra INCLUDE keyword is added for all INCLUDE statements causing an error when run with Eclipse

![image](https://github.com/equinor/subscript/assets/51825893/4cf779ae-2ebc-4669-b546-16c6d5ae29ce)

FarokhShoaei:
i have not specified the Komodo version so it is the default

mferrera:
@FarokhShoaei thanks! Just to clarify-- the default Komodo stable, or the default subscript as installed directly from here?

FarokhShoaei:
default Komodo stable

rnyb:
One can check version by running `pack_sim --version`
On current komodo stable {2023.06.06-py38} the output should look like this:  `pack_sim.py (subscript version 0.16.5)`

rnyb:
This bug appears when there is a comment line in between INCLUDE and it's argument - see example pic below
Note: when there is no comment line in between it looks ok.

![image](https://github.com/equinor/subscript/assets/51825893/6922d589-12b8-4500-9daa-0b8af87e6b0d)

rnyb:
Another thing, when re-running `pack_sim`, the output DATA file is not overwritten. We should consider changing this behavior so that it overwrites existing files. However, if written to same folder as input file then it should probably give a warning (or error) before proceeding to avoid overwriting original data file

asnyv:
A thought:

It is possible to parse an eclipse DATA-file with the opm package, and write it to a file afterwards. 
Using this would probably drastically simplify `pack_sim`, which of course is beneficial from a maintenance point of view.

The BIG disadvantages are that I think all files will be merged into one (including include files) in the export, so the DATA-file will get loooong, and I think all comments will disappear as the opm parser doesn't care about comments. 

Considering the disadvantages, I am definitely not sure that it is a good idea to make this change, but could maybe be worth a thought to minimize our technical debt in subscript.

alifbe:
> A thought:
> 
> It is possible to parse an eclipse DATA-file with the opm package, and write it to a file afterwards. Using this would probably drastically simplify `pack_sim`, which of course is beneficial from a maintenance point of view.
> 
> The BIG disadvantages are that I think all files will be merged into one (including include files) in the export, so the DATA-file will get loooong, and I think all comments will disappear as the opm parser doesn't care about comments.
> 
[...]

PR #586, opened 2023-08-11
Fix bug with extra INCLUDE keyword

alifbe:
Fix bug with extra `INCLUDE` keyword being added when there is comment between `INCLUDE`/`GDFILE`/`IMPORT` keyword and the file to be included.

Resolve issue #583

### DIFF
--- a/tests/test_pack_sim.py
+++ b/tests/test_pack_sim.py
@@ -35,6 +35,9 @@ def test_main(tmp_path, mocker):
     assert Path("include/reek.pvt").exists()
     assert Path("include/swof.inc").exists()
 
+    # Test that comment is process properly after INCLUDE keyword
+    assert "INCLUDE\nINCLUDE" not in Path(ECLCASE).read_text(encoding="utf8")
+
 
 def test_main_fmu(tmp_path, mocker):
     """Test the --fmu option on the command line, yielding
--- a/tests/test_pack_sim.py
+++ b/tests/test_pack_sim.py
@@ -35,9 +35,6 @@ def test_main(tmp_path, mocker):
     assert Path("include/reek.pvt").exists()
     assert Path("include/swof.inc").exists()
 
-    # Test that comment is process properly after INCLUDE keyword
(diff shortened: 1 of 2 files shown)


tests/test_pack_sim.py
```diff
@@ -35,6 +35,9 @@ def test_main(tmp_path, mocker):
     assert Path("include/reek.pvt").exists()
     assert Path("include/swof.inc").exists()
 
+    # Test that comment is process properly after INCLUDE keyword
+    assert "INCLUDE\nINCLUDE" not in Path(ECLCASE).read_text(encoding="utf8")
+
```
mferrera, line 40: Nice, love a simple fix 😄 however, I don't think this addition to the test catches if it fixes it or not because I don't think the reek model has the type of comment that causes the problem. Consider adding a separate regression test case for this-- something like this perhaps should do the trick (this might not be totally right):

```python
def test_double_include_before_comment(tmp_path, mocker):
     os.chdir(tmp_path)

     some_include = "some.inc"
     with open(some_include, "w", encoding="utf-8") as fout:
         fout.write("-- Comment")

     data_file = "TMP.DATA"
     with open(data_file, "w", encoding="utf-8") as fout:
         fout.write(f"INCLUDE\n--Comment\n  '{some_include}' /")
     mocker.patch("sys.argv", ["pack_sim", data_file, "out/"])
     pack_sim.main()

     # Test that comment is process properly after INCLUDE keyword
     assert "INCLUDE\nINCLUDE" not in Path(f"out/{data_file}").read_text(encoding="utf8")
```

mferrera (changes requested):

tests/test_pack_sim.py
alifbe: That's silly of me 😜 😆  . I made changes to reek model to trigger the bug, but didn't stage it. But I like your test better, so I will include it instead. 👍

mferrera (approved):
🎉👍🚀

Issue #493, opened 2022-08-16
[runrms] improvements


jcrivenaes:
When using `/prog/res/roxapi/.../matrix/releases`, there is now (from August 16, 2022) a symlink to the actual release of pythonlibs, e.g.:
 
```
11.1 -> releases/py3.6_12.0_2022.08legacy
11.1_test -> releases/py3.6_11.1_2022.08a
12.0 -> releases/py3.6_12.0_2022.08legacy
```

Here the release is e.g. `py3.6_11.1_2022.08a`

The runrms should automatically detect the release and also, as an option, display the content of the file `<release>.txt` which is a kind of `pip freeze`

In addition, when `--testpylib`, only that path shall be used as `PYTHONPATH `(now it is both)

PR #590, opened 2023-08-19
[runrms] add APS_TOOLBOX_PATH env

### LINKED ISSUE #589: [runrms] Add APS_TOOLBOX_PATH as dynamic env variable
As APS has some static code examples, a proposal is to add the APS_TOOLBOX_PATH as an env variable. This will in turn be linked to the actual PYTHONPATH

jcrivenaes:
In addition some fixed for test versions and options like --nopy

Closes #589  and address partly #493


oddvarlia (approved):
Looks OK. Works with user defined runrms.yml specification. Final test when apsgui is installed under /prog/res will have to wait untill that is done

KetilN:
Understands that setting of the APS_TOOLBOX_PATH variable should be included according to the commited code in the src/subscript/runrms/runrms.py. Will be done accordingly.

jcrivenaes:
> Understands that setting of the APS_TOOLBOX_PATH variable should be included according to the commited code in the src/subscript/runrms/runrms.py. Will be done accordingly.

Wait with that; will do in a separate PR when your structuring of `rms-sys` is finished

KetilN:
Understood. Will not do anything related to this then, but ask you when time comes.

PR #593, opened 2023-08-22
Update dependencies and CI

### LINKED ISSUE #585: Unpin sphinx
sphinx_rtd_theme has a release candidate out with support for Sphinx 7 but it's not yet on PyPI.

### LINKED ISSUE #591: Add always() to dependency installation in CI
If the linting steps always run so should the step that installs the dependencies they require.

Impetus is connection issues in the runner causing the apt installs to fail, and then jumping to the linting steps. Rare that this becomes an issue but a simple addition to make.

### LINKED ISSUE #592: Pin numpy < 2.0
NumPy 2 will arrive Dec/Jan and NumPy have recommended pinning and testing against its release candidate(s) before using. Other Equinor packages are taking this action as well.

mferrera:
Resolves #585 - sphinx-rtd-theme 1.3.0 is out and compatible with Sphinx 7
Resolves #591 
Resolves #592

### DIFF
--- a/.github/workflows/subscript.yml
+++ b/.github/workflows/subscript.yml
@@ -45,6 +45,7 @@ jobs:
           sudo apt-get install resinsight
 
       - name: Install subscript with dependencies
+        if: ${{ always() }}
         run: |
           pip install --upgrade pip
           pip install .[tests,docs]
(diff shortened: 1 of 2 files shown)


alifbe (approved):
LGTM! 👍

PR #595, opened 2023-08-23
[runrms] adapt to version format in RMS 14

jcrivenaes:


oddvarlia (approved):

oddvarlia (commented):
I tried using RMS14. But is the new APS plugin installed as a test version? 
I got aps not found. The printout from runrms was:
image

The pythonpath should include path to pydist to be able to find aps/toolbox

PR #599, opened 2023-08-28
CO2 containment script: Make containment_polygon optional argument, minor changes to calc_type_input

AudunSektnanNR:
Only minor changes.

Renaming options for calc_type_input.

Change containment_polygon from required to optional argument. If not provided (=None, the default), we will count all cells as contained.


mferrera (approved):
👍 LGTM

PR #600, opened 2023-08-30
Plume extent and plume area scripts

AudunSektnanNR:
Scripts for calculating CO2 plume area and plume extent.

These scripts are needed in the FMU for CCS project, similar to the scripts for CO2 containment calculations.

The plume area takes as input a folder of surface files (.gri) calculated using xtgeoapp_grd3dmaps. Then it calculates the area for each time step and formation, and writes the results to a CSV-file.

The plume extent script uses EGRID-file and UNRST-file to calculate the maximum extension of the CO2 plume from a given injection point.


AudunSektnanNR:
The code coverage would be much higher if we can test calling the main-methods directly with input arguments. Is this possible? Now we test methods directly, like calc_plume_extents().

mferrera:
Yep, it's possible to do this. One pattern in subscript is to patch argv with `mocker.patch()` and then call `main()`. Here's one example: https://github.com/equinor/subscript/blob/dae697e92fecf97d4effde7567347575527605d0/tests/test_eclcompress.py#L375-L377

AudunSektnanNR:
> Yep, it's possible to do this. One pattern in subscript is to patch argv with `mocker.patch()` and then call `main()`. Here's one example:
> 
> https://github.com/equinor/subscript/blob/dae697e92fecf97d4effde7567347575527605d0/tests/test_eclcompress.py#L375-L377

Thanks for answer, I will try to add a test that uses this

mferrera:
Thanks 😄 we will most likely have some time to go over this PR next week

AudunSektnanNR:
> Thanks 😄 we will most likely have some time to go over this PR next week

That's great!

mferrera:
Closing this due to the recent code organizational developments. We can reopen if something changes :-)

PR #603, opened 2023-09-18
Fix matplotlib type annotations

mferrera:
matplotlib 3.8.0 was released for Python 3.9 and above and includes type annotations in its public APIs. mypy failed on a call to `axis` and raised some notes about optional arguments.


alifbe (approved):
Nice 👍

PR #604, opened 2023-09-18
Remove protobuf as a dependency

### LINKED ISSUE #598: Double-check dependencies
`protobuf` is given as a dependency but I can't see where it's used.

We might be able to drop `packaging` as well. Maybe there are others.

mferrera:
It is pulled in through the dependency chain but is not a first class dependency.

Resolves #598


alifbe (approved):
LGTM 👍

PR #605, opened 2023-09-21
Update dependencies and linting

### LINKED ISSUE #479: [fmuobs] FutureWarning from dates in Pandas
Running `fmuobs -o yaml ert/test-data/local/snake_oil/observations/observations.txt` yields the warning:

```
/prog/res/komodo/2022.05.00-py38-rhel7/root/lib/python3.8/site-packages/subscript/fmuobs/parsers.py:414: FutureWarning: Inferring datetime64[ns] from data containing strings is deprecated and will be removed in a future version. To retain the old behavior explicitly pass Series(data, dtype={value.dtype})
  return compute_date_from_days(pd.DataFrame(obs_list), starttime)

```

mferrera:
In addition to the fixes in the commits, I spent some time trying to reproduce the below referenced issue. I tested against the old versions of pandas before and after 2022.05.00 Komodo (1.3.4 - 1.4), as well as with pandas 2, and against the historical versions of the test case observation file given as an example in the issue. All tests pass and they produce equivalent output files. But I could not reproduce the warning.

All tests pass with pandas 2 and no warnings are issued, so I think we are OK with respect to that.

Resolves #479

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,3 +1,5 @@
 [flake8]
 max-line-length = 88
 ignore = E203, W503
+per-file-ignores =
+    src/subscript/version.py:E501
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -43,7 +43,6 @@ dependencies = [
     "matplotlib",
     "numpy<2",
     "opm>=2021.10",
-    "packaging",
     "pandas",
     "pyscal",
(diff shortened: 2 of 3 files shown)


alifbe (approved):
LGTM 👍

Issue #559, opened 2023-06-22
Validate OBSDIR in MERGE_RFT_ERTOBS forward job


andreas-el:
`Hi - I just helped a user to debug crash of Ert forward job MERGE_RFT_ERTOBS  - it was a simple typo which user would have managed to fix if job exited with a proper error message.`

The issue was a typo in OBSDIR, but resulted in a cryptic error message.
If OBSDIR is found in this job, we should validate the existence for it and report back if not found.

![image](https://github.com/equinor/subscript/assets/114403625/fc59aeba-0fb6-4a51-9649-c9429b0f953c)

PR #616, opened 2023-10-11
`merge_rft_ertobs`: validate OBSDIR input 

alifbe:
Ensure that OBSDIR exist and trigger exception otherwise.

Solve #559


rnyb (approved):
LGTM

PR #625, opened 2023-10-24
`ri_wellmod` Resolve mypy issue

### LINKED ISSUE #624: mypy 1.6.1 CI failure
`src/subscript/ri_wellmod/ri_wellmod.py:246: error: Incompatible types in assignment (expression has type "Instance | None", variable has type "bool")  [assignment]`

mferrera:
Resolves #624


alifbe (approved):
@mferrera LGTM.

Just curious, Should we set default value of `resinsight`? Or will it be default to None?

mferrera:
It shouldn't be necessary since if `resinsight = rips.Instance.launch()` fails on both tries an exception will be raised, so it will never return a default value even if we set one.

mferrera:
Good point, I misread the exception handling. I updated it to function like its resinsight_dev counterpart so it explicitly returns `None` if it fails

PR #628, opened 2023-10-27
`fmuobs` Warning when using ERROR_MODE != ABS

alifbe:
ERT observation config is re-using ERROR keyword to store either absolute error value or relative error value (depending on ERROR_MODE)

However, the logic is not transferred when converting to ResInsight (csv) and probably Webviz (yml) which means if user specify 10% relative error in OBS config file. ResInsight will see it as 10 in absolute error value.

At this stage, trigger the warning when user is using ERROR_MODE that is no ABS.

The logic will be implemented in another PR.


rnyb (approved):
👍

PR #633, opened 2023-11-23
Remove co2_containment code, has been moved to new repo ccs-scripts

AudunSektnanNR:


mferrera:
Will this script reach Komodo stable from ccs-scripts from the November release, or December release?

AudunSektnanNR:
It is not included in the November release, but will be included in the December release. Currently it is only on komodo bleeding. 

We thought it would be best to remove the code from subscript in case it created some kind of conflict for the co2_containment code (subscript vs ccs-scripts)? Or is this unproblematic since it is from packages/repos with different names? Note that we haven't yet registered the code in ccs-scripts as plugin for ERT.

mferrera:
If `co2_containment` as a command line invocation exists both here and in ccs-scripts in stable, then running `co2_containment` may call it from either here _or_ ccs-scripts, but won't cause any problems otherwise. It depends mostly on the order in which Komodo installs it -- the latter one taking precedence.

So it makes a meaningful difference if co2_containment has been changed or updated in ccs-scripts, but otherwise doesn't really matter.

I guess, taking this in today, we could propose a second upgraded version for December. I think it should be okay.

mferrera:
https://github.com/equinor/subscript/blob/47f93ad28ff1c9cc27fc94bf2478ef8d6076fcaf/docs/overview.rst?plain=1#L13

Please also remove here and then I think it's good to go 😄

AudunSektnanNR:
> If `co2_containment` as a command line invocation exists both here and in ccs-scripts in stable, then running `co2_containment` may call it from either here _or_ ccs-scripts, but won't cause any problems otherwise. It depends mostly on the order in which Komodo installs it -- the latter one taking precedence.
> 
> So it makes a meaningful difference if co2_containment has been changed or updated in ccs-scripts, but otherwise doesn't really matter.
> 
> I guess, taking this in today, we could propose a second upgraded version for December. I think it should be okay.

Thanks for answer, yes there have been some new changes in ccs-scripts for co2_containment, so it will not be indentical to the version in subscript

mferrera (approved):
Unfortunately we've passed the freeze window now so these changes will come into Komodo stable in the January release.

PR #636, opened 2023-12-06
fmuobs - Add unique index to label when converting from ResInsight format

### LINKED ISSUE #635: fmuobs - Non-unique observation labels when converting from resinsight to ERT obs format
The ResInsight format does not have labels, and a label is made based on the key (vector name). This is not unique if multiple observations are present for the same key (vector), and ERT doesn't like that. 
https://github.com/equinor/subscript/blob/d0f04a1b5f73666cf6c475200de82809cf0f4385/src/subscript/fmuobs/parsers.py#L465

There is a commented part of the line which adds the date, which is an option. Another alternative is to add an index like what is done for the yaml based format:
https://github.com/equinor/subscript/blob/d0f04a1b5f73666cf6c475200de82809cf0f4385/src/subscript/fmuobs/parsers.py#L505

An index is maybe the best to be consistent with the behavior of the yaml based.

asnyv:
Resolves #635


mferrera (approved):
👍

PR #641, opened 2024-01-03
Port sunsch to use Pydantic instead of configsuite

berland:


src/subscript/sunsch/sunsch.py
```diff
@@ -102,139 +71,26 @@ def _shuffle_start_refdate(config: dict) -> dict:
...
+    template: Optional[FilePath] = None
+    days: Optional[float] = None
+    string: Optional[str] = None
+    substitute: Optional[Dict[str, Union[str, float, str]]] = None
+
+
+class SunschConfig(BaseModel):
+    files: Optional[List[FilePath]] = None
+    output: Optional[str] = "-"
+    startdate: Union[datetime.date, datetime.datetime]
```
oyvindeide, line 86: Would add: _ensure_start_refdate as a model validator, something like:
```python
    @model_validator(mode="after")
    def ensure_start_date(self) -> Self:
        if not self.startdate:
            if self.starttime:
                self.startdate = self.starttime
            elif self.refdate:
                self.startdate = self.refdate
            else:
                self.startdate = datetime.date(1900, 1, 1)
        if not self.starttime:
            self.starttime = self.startdate
       if not self.refdate:
            self.refdate = self.startdate
        return self
```
could potentially also be a pre-validation.

berland: Done, as a pre-validation. After-validation was not possible as it will invalidate based on the values given before this after_validator is run.

berland: Tried also as after-validation by setting the date fields to be Optional, allowing such an after-validator:
```
    @model_validator(mode="after")
    def set_date_defaults(self) -> "SunschConfig":
        self.startdate = (
            self.startdate
            or self.starttime
            or self.refdate
            or datetime.date(1900, 1, 1)
        )
        self.starttime = self.starttime or datetime_from_date(self.startdate)
        self.refdate = self.refdate or self.startdate
        return self
```

This was even prettier but gave uglier changes to the rest of the code as all parts have to handle the Optional fields to make mypy happy. Will stick to the before-validator.

yngve-sk: I think in general before-validators are more practical for this kind of argument inference stuff

src/subscript/sunsch/sunsch.py
```diff
@@ -724,36 +563,28 @@ def main():
...
-    if not config.valid:
-        logger.error(config.errors)
-        logger.error("Your configuration is invalid. Exiting.")
-        sys.exit(1)
+    merged_config = defaults_config.copy()
+    merged_config.update(yaml_config)
+    merged_config.update(cli_config)
 
-    if args.verbose and config.snapshot.output != __MAGIC_STDOUT__:
+    config = SunschConfig(**merged_config)
```
yngve-sk, line 570: Style: Could write something like this:

```
SunschConfig(**{
    "output": "-", 
    "startdate": datetime.date(1900, 1, 1),
    **yaml_config,
    **cli_config
})
```

instead of defining a default_config

berland: thanks, will try

yngve-sk (approved):
LGTM only some minor comments, as long the "using pre-validators for modifying the model's input dict" pattern works it is fine, would consider if overriding __init__ and doing the input parsing there instead of a validator is a more clean expression of what is being done -- filling in "missing" arguments based on other arguments given. (defaulting startdate to starttime|refdate|datetime.date(1900,1,1), defaulting starttime to startdate, defaulting refdate to startdate)... would question why that logic is there in the first place but that is maybe outside the scope of this PR.

src/subscript/sunsch/sunsch.py
berland: Pydantic did not like merging these settings; 
`E       TypeError: ModelMetaclass object got multiple values for keyword argument 'output'`

berland:
Changed from using a data validator to to the transformation to overriding `__init__`.

PR #643, opened 2024-01-05
Replace configsuite with pydantic in interp_relperm

berland:
Remove configsuite from subscripts dependencies as interp_relperm was the last user of it.


src/subscript/interp_relperm/interp_relperm.py
```diff
@@ -101,146 +100,43 @@
...
-
-
-def parse_satfunc_files(filenames: List[str]) -> pd.DataFrame:
+class Interpolator(BaseModel):
+    tables: Optional[List[int]] = []
+    param_w: Optional[Annotated[float, Field(strict=True, ge=-1, le=1)]] = None
+    param_g: Optional[Annotated[float, Field(strict=True, ge=-1, le=1)]] = None
+
+    @model_validator(mode="after")
+    def check_param_w_or_param_g(self) -> "Interpolator":
```
pinkwah, line 109: Adding `from __future__ import annotations` at the top will allow you just write `Interpolator` without quotes. This enables behaviour that will become default in some future version of Python (target used to be 3.12 but I'm not sure anymore.)

tests/test_interp_relperm.py
```diff
@@ -375,12 +332,11 @@ def test_wrong_family(tmp_path):
         "family": "Rockefeller",
         "delta_s": 0.1,
     }
-    with pytest.raises(
-        SystemExit, match="Is x a number is false on input 'Rockefeller'"
-    ):
+    with pytest.raises(ValidationError):
```
pinkwah, line 335: Should we keep the `match="Is x a number is false on input 'Rockefeller'"` from the old code, porting the error text to what Pydantic spits out of course?

src/subscript/interp_relperm/interp_relperm.py
```diff
@@ -101,146 +100,43 @@
...
+
+class InterpRelpermConfig(BaseModel):
+    low: Optional[List[FilePath]] = None
+    base: Optional[List[FilePath]] = None
+    high: Optional[List[FilePath]] = None
+    pyscalfile: Optional[FilePath] = None
+    result_file: str
+    family: Literal[1, 2] = 1
+    delta_s: float = 0.01
+    interpolations: List[Interpolator] = Field(..., min_length=1)
```
pinkwah, line 124: Style: You are using `Annotated` above, so should use it here also.

```suggestion
    interpolations: Annotated[List[Interpolator], Field(min_length=1)]
```

pinkwah (approved):
LGTM. Can't see any difference between the old and new config, and the tests seem also correct.

src/subscript/interp_relperm/interp_relperm.py
berland: fixed

src/subscript/interp_relperm/interp_relperm.py
berland: fixed.

Issue #646, opened 2024-01-10
`test_swaitinit_less_than_1_below_contact_neg_pc` fails with missing stdout.


eivindjahren:
```
def test_swatinit_less_than_1_below_contact_neg_pc(simulator, tmp_path):
        """For an oil-wet system, there can be oil below free water level.
    
        Flow will set water saturation to 1 no questions asked. Bug?
    
        Eclipse ignores SWATINIT but calculates SWAT based on the input
        Pc-curve, and can thus give SWAT<1 if pc_min < 0.
        """
        os.chdir(tmp_path)
        model = PillarModel(
            cells=1,
            apex=1000,
            owc=[900],
            swatinit=[0.7],
            swl=[0.1],
            maxpc=[3.0],
            minpc=[-3.0],
        )
        # Eclipse will pick this SWAT:
        expected_swat = 0.7915066
    
        # This must then be the Pc in the cell:
        actual_pc = -1.[610]
        p_cap = model.evaluate_pc(expected_swat)
        assert np.isclose(p_cap, actual_pc)
>       qc_frame = run_reservoir_simulator(simulator, model)

/tests/test_check_swatinit_simulators.py:531: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

simulator = 'runeclipse'
resmodel = RUNSPEC

DIMENS
  1 1 1 /

OIL
WATER

START
  1 'JAN' 2000 /

TABDIMS
  1 /

EQLDIMS
  1 /

UNIFOUT


GRID

DX
  100/
... pc@owc goc pc@goc item7 item8 oip_init
  1000 100 900 0 1* 1* 1* 1* 0/

RPTRST
  ALLPROPS/ 

SCHEDULE

TSTEP 
  1 / 


perform_qc = True

    def run_reservoir_simulator(simulator, resmodel, perform_qc=True):
        """Run the given simulator (Eclipse100 or OPM-flow)
        on a dictionary representing a dynamical reservoir model
    
        After simulation, runs check_swatinit on the results and
        returns the dataframe with QC information.
    
        Will write to cwd. Caller is responsible for starting
        in a suitable directory.
    
        If the simulator fails, the stdout and stderr will be printed.
    
        Args:
            simulator (string): Path to a working reservoir simulator
                executable
            resmodel (PillarModel): A dynamical reservoir model
[...]

PR #647, opened 2024-01-10
Safeguard retries due to license failures

berland:
Believed to solve #646


eivindjahren (approved):
Necessary anyways, but there may still be an underlying issue.

PR #650, opened 2024-01-22
Fix licence retry behaviour

berland:
stderr=subprocess.PIPE was missing which implied that result.stderr was always None. This tripped the license retry code to never retry. Also when this is set correctly, result.stdout and result.stderr are never None and need not to be tested for.


jonathan-eq (approved):
LGTM!

PR #654, opened 2024-01-29
Lint with ruff

mferrera:


src/subscript/csv2ofmvol/csv2ofmvol.py
```diff
@@ -71,7 +70,7 @@
 
 
 def read_pdm_csv_files(
-    csvfiles: Union[pd.DataFrame, str, List[str], List[pd.DataFrame]]
+    csvfiles: Union[pd.DataFrame, str, List[str], List[pd.DataFrame]],
```
alifbe, line 73: @mferrera interesting that ruff add comma here. Do you know why?

alifbe (approved):
Looks good to me 👍

src/subscript/csv2ofmvol/csv2ofmvol.py
mferrera: If I had to guess, it's just because it's easier to add things afterward... and also to indicate to formatters that this should indeed by a multi-line list and not inline

PR #666, opened 2024-02-21
Fix bug for integers in sunsch substitute

### LINKED ISSUE #665: `sunsch` substitute feature converts `int` to `float` after #643

asnyv:
Resolves #665 

The issue was that certain inputs have to be `int` to be parsed by `opm`, e.g. the VFP table number (added it to the test). 
Probably just a typo in the conversion to `pydantic` as `str` was repeated twice in the typing.


asnyv:
This is a bug in release `1.2.0`, should a patch `1.2.1` be made?

PR #670, opened 2024-02-26
Change to standard column names in `params2csv`

### LINKED ISSUE #668: Standardize column names in `params2csv` output
Current issues with  `params2csv`

- The output csv when run on an ensemble (from CLI or ERT WORKFLOW) has non-standard column names `Realization` and `Iter`. 
- The output csv does not include an `ENSEMBLE` column containing the name of the iteration folder e.g. "iter-0", "iter-1", or "pred". Only the `filecolumnname` keep an indirect record of this now.
- The output csv does not include an `ENSEMBLESET` column containing the name of the case folder.
- when dropping constant parameters, also the case metadata columns like `Realization` and `Iter` is dropped if no variation.

Per now the only way to get a standard csv-file from an ensemble using  `params2csv` is by running it as a FORWARD_MODEL in ERT, and subsequently merging the individual csv-files with `CSV_MERGE`.
(`Realization` and `Iter` columns will not be present in the csv-file outputted per realization, and  `CSV_MERGE` will add the  `ENSEMBLESET`, `ENSEMBLE` and `REAL` columns.)

The column names should be changed to the more standard FMU names `ENSEMBLESET`, `ENSEMBLE`, `REAL` and `ITER` when running on an ensemble,  so that it aligns with the csv-file you would get by using the FORWARD_MODEL and CSV_MERGE.

tnatt:
PR to standardize the output csv from `params2csv` when run on an ensemble. Resolves #668 

- Extract case metadata into standard column names `ENSEMBLESET`, `ENSEMBLE`, `REAL` and `ITER` (similar as `csv_merge`).
- Removed the old  `Realization` and `Iter` columns.
- Changed to only dropping constant **parameters columns** 

  
**Note**, it was considered keep and deprecate the `Realization` and `Iter` columns for a period. Still not sure if this is needed.. Discussed with @alifbe and @rnyb, and we all tend to lean towards dropping these columns.
- this tool is not expected to be very much in use,  it is mainly used as an post-processing step 
  - The ERT WORKFLOW functionality is very new (just a couple of months) and has not been added to the Drogon dataset yet. Hence few people have probably picked up this functionality. 
- the change is not breaking any logic, and it should be an easy fix for the users. 

Any thought on this  @mferrera

### DIFF
--- a/src/subscript/params2csv/params2csv.py
+++ b/src/subscript/params2csv/params2csv.py
@@ -4,11 +4,13 @@ data, ensuring labels for each value matches).
 
 """
 
+from __future__ import annotations
+
 import argparse
 import logging
-import re
 import shutil
 from glob import glob
+from pathlib import Path
 
 import pandas as pd
 from ert.config import ErtScript
@@ -27,7 +29,7 @@ parameters.txt is a text file with <key> <value> on each line
 In the CSV file, each individual parameter file will be represented by one data row.
 The order of parameters in each text file is not conserved.
 
-The original filename for each file is written to the column ‘filename’.
+The original filename for each file is written to the column 'filename'.
 Beware if you have that as a <key> in the text files.
 """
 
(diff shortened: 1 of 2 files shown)


src/subscript/params2csv/params2csv.py
```diff
@@ -153,26 +154,33 @@ def params2csv_main(args: argparse.Namespace) -> None:
...
+    for parameterfilename in paramfile_paths:
+        if parameterfilename.exists():
+            paramtable = pd.read_csv(
+                parameterfilename,
+                names=["key", "value"],
+                header=None,
+                usecols=[0, 1],
+                sep=r"\s+",
+            )
+        else:
```
mferrera, line 180:
```suggestion
        if not parameterfilename.exists():
            logger.warning("%s not found, skipping..", parameterfilename)
            continue
            
        paramtable = pd.read_csv(
            parameterfilename,
            names=["key", "value"],
            header=None,
            usecols=[0, 1],
            sep=r"\s+",
        )
```

more of a stylistic nit, not really important (this suggestion doesn't contain the log warning just below it)

tests/test_params2csv.py
```diff
@@ -165,28 +174,124 @@ def test_ert_workflow(tmp_path):
             f"real\t{i}", encoding="utf8"
         )
 
-    Path("PARAMS2CSV_ITER0").write_text(
+    Path("PARAMS2CSV").write_text(
```
mferrera, line 177: I think you might actually want to keep this different from `PARAMS2CSV` since that is the forward model name. Maybe it doesn't make a difference... `PARAMS2CSV_ITER0` is what is suggested in the documentation text too.

mferrera (commented):
This seems reasonable to me. I think Alif or Roger should probably stamp an approval over myself.

tests/test_params2csv.py
tnatt: I think my reasoning was that most users probably want to extract one csv-file for all iterations, and not one per iteration.  At least now after standardizing the csv-file and adding more case metadata in the output. But the tests are not where most users go to see how to use this tool anyway.. 😄 
Maybe we should update the ERT WORKFLOW documentation to show how to extract all iterations in one go to the users instead? As the documentation is today, many would probably think that they would have to have multiple ERT WORKFLOWS to extract all the iterations. Agree @alifbe @rnyb ?

src/subscript/params2csv/params2csv.py
tnatt: Yes, agree! I like this better 👍

tests/test_params2csv.py
rnyb: I don't have a strong opinion on this, but can it be potentially be a problem if for instance mixing of hist and pred?

rnyb: Anyways, I'll guess the user can still choose however they want to use it

rnyb (approved):
LGTM

mferrera:
Please update the commit message before merging 👍

tnatt:
As discussed with @alifbe and @rnyb: 
I've extended the ERT example with an example showing how to extract parameters for all iterations in one go. In addition I've  changed the filename for the ERT workflow file to lower case and added "wf_" as prefix to align with standards in Drogon 👍

Issue #669, opened 2024-02-26
[check_swatinit] Tests fail on RHEL8 builds


andreas-el:
Update:

The tests are now skipped instead:

```
SKIPPED [2] tests/test_check_swatinit_simulators.py:287: Test require flow dev version
SKIPPED [2] tests/test_check_swatinit_simulators.py:338: Test require flow dev version
SKIPPED [2] tests/test_check_swatinit_simulators.py:482: Test require flow dev version
SKIPPED [2] tests/test_check_swatinit_simulators.py:573: Test require flow dev version
SKIPPED [2] tests/test_check_swatinit_simulators.py:606: Test require flow dev version
```

Subscript tests for check_swatinit fails during komodo bleeding builds using python `3.8`:
https://github.com/equinor/komodo-releases/actions/runs/8041838251/job/21963435048

As a sidenote, these also fails for python version `3.11`
https://github.com/equinor/komodo-releases/actions/runs/8041838251/job/21962580908

```
FAILED tests/test_check_swatinit_simulators.py::test_swatinit_1_far_above_contact[flow] - assert False
 +  where False = <function isclose at 0x7f2ae5d8a700>(0.1, 1)
 +    where <function isclose at 0x7f2ae5d8a700> = np.isclose
FAILED tests/test_check_swatinit_simulators.py::test_swatinit_1_slightly_above_contact[flow] - assert False
 +  where False = <function isclose at 0x7f2ae5d8a700>(0.88782424, 1)
 +    where <function isclose at 0x7f2ae5d8a700> = np.isclose
FAILED tests/test_check_swatinit_simulators.py::test_swatinit_less_than_1_below_contact_neg_pc[flow] - assert False
 +  where False = <function isclose at 0x7f2ae5d8a700>(0.7906856, 1.0)
 +    where <function isclose at 0x7f2ae5d8a700> = np.isclose
FAILED tests/test_check_swatinit_simulators.py::test_swu_equal_swatinit[flow] - assert False
 +  where False = <function isclose at 0x7f2ae5d8a700>(0.5152699, 0.9)
 +    where <function isclose at 0x7f2ae5d8a700> = np.isclose
FAILED tests/test_check_swatinit_simulators.py::test_swu_lessthan_swatinit[flow] - assert False
 +  where False = <function isclose at 0x7f2ae5d8a700>(0.46336114, 0.9)
 +    where <function isclose at 0x7f2ae5d8a700> = np.isclose
[...]

PR #672, opened 2024-02-28
Update test for check_swatinit

alifbe:
`OPM flow` has an update in saturation modelling which affect the test for `check_swatinit`

Solve #669 

The PR is doing two things:
- Update test to reflect the new behavior of OPM flow
- Skip some of the test when run from GitHub because the change is not yet in official `OPM flow`


rnyb (approved):
LGTM

Issue #601, opened 2023-09-11
Remove runrms


mferrera:
It should be easier to manage the transition from the runrms script here to the new version being matured in rms-sys if they exist in the same repository.

- [x] Move `runrms` and its tests to rms-sys
- [x] Rename `runrms` to `runrms_legacy`
- [ ] Remove command line invocation from `pyproject.toml`

PR #684, opened 2024-04-09
Rename `runrms` cli invocation

mferrera:
This may be temporary or permanent until rms-sys takes precedence. Renaming this for komodo bleeding so that rms-sys' `runrms` may be tested more simply from Komodo bleeding and we can get a clearer idea if anything else is malfunctioning or relied upon certain things tied to this specific command line invocation.

There's a risk we forget to undo this. 
- #685 
- #601


alifbe (approved):
LGTM

Issue #686, opened 2024-04-15
bjobsusers will not work on RHEL8


berland:
bjobsusers use `finger` to map shortnames to full names. The finger utility is removed in RHEL8.

https://github.com/equinor/subscript/blob/a1ddb061ee136035d76b9c2bc9a946c39678d146/src/subscript/bjobsusers/bjobsusers.py#L80

`pinky -l <username>` can be used as a replacement.

PR #688, opened 2024-04-16
Replace finger with pinky

alifbe:
Solve #686


mferrera (approved):
This look okay to me.

PR #689, opened 2024-04-17
Included a few commonly seen Eclipse/Flow keywords

mvasdan:
I have included a few commonly seen Eclipse/Flow keywords that are commonly seen in most of the simulation models.


mferrera (approved):
Thanks very much for making this list more complete.

PR #697, opened 2024-05-09
Update repo for opm

alifbe:
There is an update on how to install opm flow

https://opm-project.org/?p=1721

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -26,7 +26,7 @@ jobs:
     - name: Install OPM-flow and ResInsight (for testing)
       run: |
         sudo apt-get install software-properties-common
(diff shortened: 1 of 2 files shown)


.github/workflows/codecov.yml
```diff
@@ -27,6 +27,7 @@ jobs:
       run: |
         sudo apt-get install software-properties-common
         sudo apt-add-repository ppa:opm/ppa
```
rnyb, line 29: Should this still be included?

alifbe: Yes, I removed it in my first commit. But `resinsight` is is install from that repo `ppa:opm/ppa`

rnyb (approved):
LGTM

PR #703, opened 2024-06-04
Grav subs maps

tralsos:
Added job for modelling of gravity and subsidence maps from flow simulation results.


src/subscript/grav_subs_maps/grav_subs_maps.py~
alifbe, line 1: You can probably delete this file

tralsos: yes, will do

tests/testdata_gravity/grav_subs_points.yml
```diff
@@ -0,0 +1,16 @@
+# Configuration file for use with grav_subs_points.py
+
+input:
+  diffdates:
+    - [2020-07-01, 2018-01-01]
+    
+stations:
+  grav:
+    '2020_2018': /project/fmu/users/tral/subscript/subscript/tests/testdata_gravity/station_coordinates.csv
```
alifbe, line 9: I guess this configuration file has not been used for testing. But still it will be good to keep it as relative path.

alifbe (approved):
This looks good to me. 👍 

Not sure if there is a use case for running the script on CLI, but I think it will be great if more detail documentation can be provided, in addition to what you have made for ERT documentations (`DESCRIPTION` and `EXAMPLES`)

As example, you can check:
-  https://github.com/equinor/subscript/blob/9f70d93fe42a014bf336b0e1127b8a3c5d8f2a0d/docs/scripts/check_swatinit.rst
- https://github.com/equinor/subscript/blob/9f70d93fe42a014bf336b0e1127b8a3c5d8f2a0d/docs/scripts/casegen_upcars.rst

tests/testdata_gravity/grav_subs_points.yml
tralsos: You are right, this config was added to the test dataset, but will be used in a job not implemented yet. But will update to relative path now.

tralsos:
Thanks @alifbe. Will update with a more detailed documentation after adding the other gravity jobs.

src/subscript/grav_subs_maps/grav_subs_maps.py
```diff
@@ -0,0 +1,324 @@
...
+                    )
+                    sys.exit(1)
+    phase_code = {"oil": 1, "gas": 2, "water": 4, "total": 7}
+
+    # Gravity
+    for diffdate in diffdates:
+        for phase in phases:
+            logger.info(
+                f"Calculating delta gravity map from {phase} "
+                "for {diffdate[0]}_{diffdate[1]}"
```
alifbe, line 268: I think you need to add `f` here otherwise it won't be expanded.

src/subscript/grav_subs_maps/grav_subs_maps.py
```diff
@@ -0,0 +1,324 @@
...
+            rsb = rest.restartView(0)
+            if singledate not in added_dates:
+                if singledate in restart_index:
+                    rsb = rest.restartView(restart_index[singledate])
+                    grav.add_survey_RFIP(singledate, rsb)
+                    subsidence.add_survey_PRESSURE(singledate, rsb)
+                    added_dates.append(singledate)
+                else:
+                    logger.error(
+                        f"Date {singledate} specified but not" "found in UNRST file."
```
alifbe, line 258: Small typos? Add space between `not` and `found`

PR #704, opened 2024-06-04
Add Genertobs

daniel-sol:
Solving issue equinor/ert#10795. Add module for generation of observation data for ert from yaml config file that takes csv, or spreadsheets as input. 

Extra feature added: Users can deactivate observations, either in the config file to remove the entire observation group:

```yaml
- name: Gas-oil-ratio observations
  type: timeseries
  error: 10%  # optional (% indicates relative, otherwise absolute)
  active: false
  observation: ./input/observations/summary_gor.csv
```
Or directly in the csv file, by adding an active column.
This column is not required, and when not present all observations are included
When present only the entries that do not include a **no** will be included
| value  | error |active|
| ------- | ----- |----- |
| 0.25  | 0.1  |  |
| 0.23 | 0.09  |  |
| ..|..|..|
| ..|..|..|
|0.32|0.15|no|
|0.33|0.15|no|
| ..|..|..|

The name active has been by several suggested to be suboptimal, any suggestion for a different name would be appreciated


mferrera:
I'm uncertain that the upload workflow should live in this repository. Architecturally it feels more appropriate for that functionality to live in the Sumo world, particularly if it's importing private classes and functions from `fmu-sumo-uploader` anyway. Additionally, I don't think we want `fmu-sumo-uploader@git+https://github.com/equinor/fmu-sumo-uploader"` (bleeding) as a dependency here. 

Have you considered locating the uploader there?

daniel-sol:
> I'm uncertain that the upload workflow should live in this repository. Architecturally it feels more appropriate for that functionality to live in the Sumo world, particularly if it's importing private classes and functions from `fmu-sumo-uploader` anyway. Additionally, I don't think we want `fmu-sumo-uploader@git+https://github.com/equinor/fmu-sumo-uploader` (bleeding) as a dependency here.
> 
> Have you considered locating the uploader there?

I think this a fair point, I was just very focused on making it work, therefore I added the uploader as a dependency so that one could test it during development. But I think it makes to generate the include file if the cli script upload_obs_2_sumo (or upload_preprocessed_2_sumo?) Already exists? 
So suggested action plan will then be:

- Split off dataio and sumo bit, relocate to uploader?
- Split off generation of workflow part to separate PR after the relocation is merged in

src/subscript/genertobs_unstable/_datatypes.py
```diff
@@ -0,0 +1,325 @@
...
+        bool: check passed or not
+    """
+    try:
+        float(tocheck)
+        return True
+    except TypeError:
+        return False
+
+
+def is_percent_range(string):
```
alifbe, line 36: This doesn't seem to check if the string contains %

alifbe: But I see that the checking is done before calling the function. Could probably update the docstring.

PR #707, opened 2024-06-07
`prtvol2csv` test again latest simulator output

alifbe:
- Add Drogon dataset
- Refactor `run_reservoir_simulator` from `test_check_swat_init_simulator` so that it can be used in `test_prtvol2csv`
- Ensure that PRT file from latest version of simulator is still supported.


rnyb (approved):
LGTM

PR #718, opened 2024-07-24
`runeclipse`: Support multiple license servers

### LINKED ISSUE #714: Need support for multiple license servers + local license servers on some sites
We need that `runeclipse` sets up its license server environment variable differently

This is part of stabilizing / improving license server responses. `runeclipse` as it is now – used from remote sites - causes response problems for license services in Stavanger

We need 3 changes – basically:

1. Use two license server names in `SLBSLS_LICENSE_FILE` environment variable– random order – separated by : (colon)
     - _eclipse1-lic-no.statoil.no_
     - _eclipse2-lic-no.statoil.no_
4. StJohn and Rio now have local licenses for Eclipse – so _stjohn-grid01.stjohn.statoil.no_ or _rio-grid01.rio.statoil.no_ is prepended there
5. Using the `LM_LICENSE_FILE` environment variable is obsolete – please use `SLBSLS_LICENSE_FILE` instead

You can use `/prog/res/ecl/script/eclrun-2x.bash` or `/prog/res/ecl/script/eclrun-2x.cshrc` as a prototype / demonstration for what we have tested to work with eclrun
 
You should be able to do this at any time. Both names + _eclipse-lic-no.statoil.no_ all points to _st-vlic20.st.statoil.no_

alifbe:
Close #714

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ dynamic = ["version"]
 dependencies = [
     "resdata",
     "res2df",
-    "ert>=2.38.0b7",
+    "ert",
     "fmu-tools",
     "matplotlib",
     "numpy<2",
--- a/src/subscript/legacy/runeclipse
+++ b/src/subscript/legacy/runeclipse
@@ -8,14 +8,25 @@ use Cwd;
 use IO::File;
 
 # ------------------------------------------------------
+my $hostname           = `hostname`;
+my $_balance     = `date +%s`;
 
+$_balance = $_balance % 2;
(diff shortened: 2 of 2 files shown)


mferrera:
I'd suggest removing the ert bottom pin in pyproject.toml for this PR and we can re-add it in a follow-up PR, after making a new release. I guess there is value in having subscript run against ert pre-release versions in the nightly ci here.

mferrera (approved):
This looks OK to me. I have not tested it. 👍

PR #720, opened 2024-07-30
Use opm TimeVector in sunsch

asnyv:
Resolves https://github.com/equinor/subscript/issues/467

This was afaik resolved upstream by https://github.com/OPM/opm-common/pull/3188
which has been included in `opm>=2023.04`, hence the bumped requirement.

Nice if @vkip can confirm this prior to merging.

PR #722, opened 2024-08-16
Use sunsch updated time_vector.py

alifbe:
Use @vkip updated time_vector.py to be compatible with opm 2024.4

https://github.com/OPM/opm-common/pull/4165

### DIFF
--- a/src/subscript/sunsch/sunsch.py
+++ b/src/subscript/sunsch/sunsch.py
@@ -16,10 +16,10 @@ from typing import Dict, List, Literal, Optional, Union
 
 import dateutil.parser
 import yaml
-from opm.tools.time_vector import TimeVector
 from pydantic import BaseModel, FilePath
 
 from subscript import __version__, getLogger
+from subscript.sunsch.time_vector import TimeVector  # type: ignore
 
 logger = getLogger(__name__)
 
--- a/src/subscript/sunsch/time_vector.py
+++ b/src/subscript/sunsch/time_vector.py
@@ -1,0 +1,425 @@
+# To be removed once the OPM version of this file is updated
+
+import datetime
+from operator import attrgetter
+
+try:
+    from StringIO import StringIO
+except ImportError:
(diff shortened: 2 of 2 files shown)


mferrera (approved):

asnyv:
We just removed the copied code from last time there was a fix in opm (https://github.com/equinor/subscript/pull/720). I really think we should try to avoid having this code in subscript. Either push for a patch release in opm-common or wait for the next (as far as I know we have lived without the functionality in https://github.com/OPM/opm-common/pull/4165 for a long time.

alifbe:
> We just removed the copied code from last time there was a fix in opm (#720). I really think we should try to avoid having this code in subscript. Either push for a patch release in opm-common or wait for the next (as far as I know we have lived without the functionality in [OPM/opm-common#4165](https://github.com/OPM/opm-common/pull/4165) for a long time.

@asnyv Yes, I totally agreed with you. But we got a report where user has crash when they have `BRANPROP` keyword in schedule section due to the changes in OPM 2024.4 
https://equinor.slack.com/archives/CKC3ST0UT/p1723631464314409

We will revert this once PR in OPM is released (probably in 2024.10)

asnyv:
@alifbe alternatively pin opm to 2023.10 until the fix is in place in 2024.10?

mferrera:
Seeking to pin opm for the August Komodo was part of the plan as well. In our discussion about this, and trying to account for all the moving parts with RHEL8 etc, we thought reverting this change would be needed too. But I think now, maybe it is not?

asnyv:
I'm not going to completely block this PR of course but I would at least recommend to check if a pin is sufficient first.

mferrera:
For posterity, we decided to merge this as it contains an additional patch fix going into OPM's time_vector.py as well, and pinning OPM < 2024.4 interferes will cause us to lack RHEL8 compatibility. We will revert the change when 2024.10 comes out

PR #724, opened 2024-08-19
job for gravity and subsidence modelling at points

tralsos:
Added new forward model for modelling of gravity change and subsidence at point locations at seabed (benchmark stations), based on flow simulation results. Uses a yml config as input.


tests/test_grav_subs_points.py
alifbe, line 1: Would it make sense to also add test this as ERT forward model?

alifbe (approved):
LGTM 👍

Issue #726, opened 2024-08-21
Update examples with new variable names


tralsos:
Examples for running grav_subs_points and grav_subs_maps with ert needs update after change of config.

PR #727, opened 2024-08-21
Update examples

tralsos:
Update of examples of how to configure with ert. Change in name of some variables, and adding more examples.


tralsos:
Solving #726 .  Change only in examples, not in code or tests.

mferrera (approved):
👍

Issue #736, opened 2024-09-10
remove root path option 


tralsos:
grav_subs_maps and grav_subs_points currently has an option in configuration to specify which root path that should be used for file names in the yaml configuration. This was an idea inherited from the interp_relperm job. However, in this case this will probably never be used, and complicates both code and configuration.

Probably better to instead use similar thinking as for the grid3d_average type of jobs where the root path is where the job is run from, and any relative paths in the yaml config file should be relative to this.

Issue #728, opened 2024-08-21
gravity: read dates from included file


tralsos:
Currently the restart dates to use in grav_subs_points and grav_subs_maps are specified directly in the yaml configuration file for the job. It would be better if we could include dates from a global_variables file, where the same dates are also used for other calculations. This is similar to what is done for configuration of grid3d_average_maps.

PR #737, opened 2024-09-11
Remove root option and add include option for dates

tralsos:
Solves #736 and #728


tralsos:
Removing this and creating new pull request from another branch

PR #739, opened 2024-09-13
Add script to calculate mean, stdev for continuous 3D field parameter and facies probabilities for 3D facies parameter per zone

oddvarlia:
…s and estimated facies probabilities for discrete 3D parameter


oddvarlia:
Purpose:
Calculate ensemble mean and standard deviation for 3D continuous parameters and estimates for facies probabilties for 3D facies parameters.
The estimates should be available per zone in the geomodel grid also for ensembles where the 3D geomodel grid vary from realization to realization as long as number of grid cells laterally (nx, ny) is fixed. To handle this, a fixed "gox" grid called ERTBOX grid is used.
The plan is also to be able to calculate this for field parameters used in APS in AHM workflows in ERT using ES-MDA:

oddvarlia:
I will add a specification using yaml as config file format instead of directly specifying a python dict.

pyproject.toml
alifbe, line 1: I think you also need to register the workflow job in `[project.entry-points.ert]` section.

src/subscript/field_statistics/FIELD_STATISTICS
alifbe, line 1: I believe you don't need this file with the way it is set-up

src/subscript/field_statistics/field_param_stat.yml
alifbe, line 1: Is this file still needed?

src/subscript/field_statistics/wf_field_statistics
alifbe, line 1: I believe this file is not needed anymore?

tests/testdata_field_statistics/ert/model/0readme
alifbe, line 1: Does the script required this folder structure? Is that the reason for having this 0readme file?

alifbe:
I think this PR looks great. I am still not sure if you need to have `ert/model` in your testdata folder. 

Perhaps @rnyb  or @mferrera  can also have a look at this PR?

pyproject.toml
oddvarlia: Done

src/subscript/field_statistics/FIELD_STATISTICS
oddvarlia: Is now removed

src/subscript/field_statistics/field_param_stat.yml
oddvarlia: Is now removed

src/subscript/field_statistics/wf_field_statistics
oddvarlia: Is now removed

tests/testdata_field_statistics/ert/model/0readme
oddvarlia: The command line input take the CONFIG_PATH for ERT as one of the input, and the code read the global_variables.yml file from fmuconfig/output to get realization independent info about facies per zone and for that purpose I use the ERT CONFIG PATH and therefore use this directory structure in the testdata

mferrera (approved):
There are a few improvements I think are possible here (print statements, commented out code, etc) but as a first pass to initialize a script I think we can take this as is for now.

Issue #740, opened 2024-09-13
workaround for input from opm flow


tralsos:
The grav_subs jobs are using resdata gravity module for calculations. The recommended method is add_suvery_RFIP which uses RFIP for each phase. However RFIP is currently not supported by opm flow, and is missing in the unrst files. 

A possible workaround is to use the method add_survey_RPORV for cases where RFIP is missing. In parallel adding RFIP in flow is added in their backlog.

PR #743, opened 2024-09-26
add rporv workaround

tralsos:
When RFIP for each phase is missing in the UNRST file, the preferred resdata gravimetry method add_survey_RFIP is failing. A workaround is to use the add_survey_RPORV method instead, which calculates RFIP from RPORV and phase densitities. This is currently needed when using OPM FLOW as simulator, since it does not output RFIP to the UNRST file.  Solves #740 

Adding this workaround allows for implementation of the gravity modelling in Drogon.


alifbe (approved):
Looks good to me 👍

PR #744, opened 2024-09-27
Added  a file to export and copy result files if specified to realization-0 for use with covizualization in RMS

oddvarlia:
Small fixes related to alternative directories for results and include a copy of the ERTBOX.roff file


oddvarlia:
This branch is identical to the one in PR 739 except that additional option for where to put the result is added and also that ERTBOX.roff grid is added to the result directory. The fix was done to make it easier to use RMS covizualization  to visualize the results.

mferrera (approved):
OK, you'll have to rebase into this one

PR #746, opened 2024-10-01
Fixed output file names from the generated rms script

oddvarlia:
Some parameters loaded into rms using the generated rms script from field_statistics.py lack the user defined label. This is fixed here.


alifbe (approved):
Looks good to me 👍

PR #749, opened 2024-10-07
Copy mean, stdev and estimated facies probabilities to one realization of the geogrid

oddvarlia:
The purpose of the new option -g is to copy the results (mean, stdev, facies prob estimates) from ertbox grid which is fixed and only works as a "container" for the values into a single realization (realization 0) of the geogrid. The argument for this is to be able to visualize the results together with blocked well facies log and petrophysical logs.

There are several things to notice here:

- Since the result is copied into a single grid realization, only the results for grid cells for that grid realization will be  possible to see. But if we assume that the grid realization is based on a structural model conditioned to the wells, it should be possible to inspect the results along the well paths in the single realization.
- The statistical results are based on the main assumption that field values (facies, petrophysical values) belonging to the same grid cell index belongs to the same probability distribution regardless of the whether the geometry of the geogrid varies laterally or vertically from realization to realization. 
- Since number of active grid cells vary from realization to realization, the confidence of the estimates of mean, stdev, facies probabilities will vary from grid cell to grid cell depending on how many realizations have the given grid cell as an active grid cell. Hence, for top or base conform grids this will vary a lot at the base or top of the zone. For grids with stair case faults, also along these faults the active grid cells will vary a lot and the number of realization for each grid cell index will vary from grid cell to grid cell around these faults.


src/subscript/field_statistics/field_statistics.py
```diff
@@ -68,16 +68,56 @@
 specified folder, but default if not specified is share/grid_statistics
 folder under the top level of the scratch directory for the ERT case.
 The default estimate of standard deviation is the sample standard deviation
-(variance = sum( (x(i)-x_mean)^2 )/(N-1) )
+
```
jcrivenaes, line 71: You can use the math directive here to render formulas (~latex'ish):

```
.. math::

   \\text{variance} = \\frac{\\sum (x_i - \\bar{x})^2}{N-1}

```

src/subscript/field_statistics/field_statistics.py
```diff
@@ -163,26 +203,41 @@
...
+of the geogrid under realization-0 in the directories:
+
+    realization-0/iter-0/share/results/grids/geogrid--<resultparameter>.roff
+    realization-0/iter-3/share/results/grids/geogrid--<resultparameter>.roff
+
+where iter-0 is initial ensemble and iter-3 is final ensemble.
+The default (when this option is not specified) is not to copy results into
+one realization of the geogrid.
+
+Workflow job for ERT to calculate field statistics FIELD_STATISTICS
```
jcrivenaes, line 237: As discussed off-line, I think it is important to state in the documentation that the averaging has weknesses when vertical number of grid cells are varying for realization to realization. State somehow that it is "approximative".

src/subscript/field_statistics/field_statistics.py
```diff
@@ -981,37 +1277,54 @@ def main():
...
-                            "ertbox--" + stat + "_" + zone + "_" + prop_name \
-                            + "_" + str(iteration)
+    for zone in zone_list:
+        if cont_prop_dict:
+            if zone in cont_prop_dict:
+                for stat in stat_list:
+                    for prop_name in cont_prop_dict[zone]:
+                        for iteration in iter_list:
+                            name =  \
+                            "ertbox--" + stat + "_" + zone \
```
jcrivenaes, line 1298: As a future todo: use `fmu-dataio` to export these. Currently, these results will not be availbale in SUMO.

jcrivenaes (approved):
A few comments to consider, otherwise LGTM

src/subscript/field_statistics/field_statistics.py
oddvarlia: OK. Tell me how to do that offline..

src/subscript/field_statistics/field_statistics.py
oddvarlia: Agree.

alifbe (approved):
LGTM 👍

PR #752, opened 2024-10-11
Some documentation reformatting for ERT workflow script FIELD_STATISTICS

oddvarlia:
Fixed some formatting issues related to the documentation of this script appearing in ERT documentation.

### DIFF
--- a/src/subscript/field_statistics/field_statistics.py
+++ b/src/subscript/field_statistics/field_statistics.py
@@ -28,11 +28,16 @@ from ert.config import ErtScript
 import subscript
 
 logger = subscript.getLogger(__name__)
+DESCRIPTION_FOR_ERT = """Calculate mean, stdev and estimated facies probabilities
+from field parameters using ERTBOX grid.
+"""
+
 DESCRIPTION = """Calculate mean, stdev and estimated facies probabilities
 from field parameters using ERTBOX grid.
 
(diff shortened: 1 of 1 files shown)


alifbe (commented):
@oddvarlia Is there any specific reason to have different parser for ERT and CLI? I just afraid that the maintenance will be more cumbersome.

oddvarlia:
I had problem with the ERT GUI and the documentation coming twice there. The only thing I did was not to change the parser, but to change the documentation to a short version for ERT, but in the code there are two different functions with only the text for the documentation that are different. I see that for some other scripts in subscript repo, there are also two versions of the parser where lots of optional settings were left out in the parser version for ERT which then used default settings for those options, while the subscript version had all options available. But if you have a better way to fix the documentation issue in the ERT version we can use that instead.

alifbe (commented):
I see that `csv_merge` is also doing the same thing. 😄 

I guess it's useful if you want to hide some of the arguments from ERT documentation, otherwise `argparse` will set it to default anyway if it's not given. The choice is yours 😄 

Regarding duplicate of description, you can try to remove the description in line 1488. Something like: `workflow.parser.description = ""` will reset the description of the parser and hence remove duplicate text.

alifbe (approved):
LGTM 👍 

I added some comments to consider.

PR #756, opened 2024-10-18
Check that the ensemble exists for the specified iterations before calculating anything

oddvarlia:
Purpose: Add a check that the ensemble on scratch disk has completed with the iterations specified in the config file for this script.
This is a workaround to ensure that the calculations of field parameter statistics are started before the final realizations are created. Using the HOOK_WORKFLOW with POST_SIMULATION will start the workflow after each iteration, not after all iterations are completed. This will trigger unnecessary runs of the calculation of field parameter statistics since the only meaningfull calculation is after the final iteration is completed. By adding this check, it is possible to avoid unnecessary calculations and waste of time. Ideally ERT should have an option for HOOK_WORKFLOW that start a workflow in ERT after all iterations are completed, but that is not available by the time of writing.

### DIFF
--- a/src/subscript/field_statistics/field_statistics.py
+++ b/src/subscript/field_statistics/field_statistics.py
@@ -135,6 +135,10 @@ of the well conditioning since the blocked wells grid cells may vary
 from realization to realization due to the variability of the grid which is
 due to structural uncertainty.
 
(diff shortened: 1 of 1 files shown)


alifbe (approved):
LGTM 👍

PR #757, opened 2024-10-24
Mean and stdev of intermediate field parameters

oddvarlia:
Purpose:
For QC purpose it could be useful to check mean and stdev also for intermediate/temporary field parameters that are saved under <RUN_PATH>/rms/output/aps directory for initial ensemble field parameters and on <RUN_PATH> for updated field parameters. Here the GRF's for APS and possibly also transformed gaussian fields for petrophysical parameters are saved per facies per zone.
This can be specified by some additional keywords in the input config file for the field_statistics.py script

### DIFF
--- a/src/subscript/field_statistics/field_statistics.py
+++ b/src/subscript/field_statistics/field_statistics.py
@@ -35,9 +35,16 @@ from field parameters using ERTBOX grid.
 DESCRIPTION = """Calculate mean, stdev and estimated facies probabilities
 from field parameters using ERTBOX grid.
 
-The script reads ensembles of realizations from scratch disk  from::
+The script reads ensembles of realizations from scratch disk  from
+<RUN_PATH> directory::
 
   share/results/grids/geogrid--<propertyname>.roff.
+
+Optionally also temporary field parameters coming from APS or petrophysical
+field parameters can be used to calculate mean and standard deviations.
+The ensemble of realizations are usually located under <RUN_PATH> directory::
+
+  rms/output/aps
 
 Since the realizations may have a grid geometry that is realization dependent
 and may have multiple zones, the values are first copied over to a static grid
@@ -145,8 +152,7 @@ but return without doing anything.
 EPILOGUE = """
 .. code-block:: yaml
 
-  # Example config file for wf_field_param_statistics
-
+  # Configuration file for script wf_field_param_statistics.py
   field_stat:
     # Number of realizations for specified ensemble
     # Required.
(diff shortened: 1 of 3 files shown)


src/subscript/field_statistics/field_statistics.py
```diff
@@ -314,6 +359,13 @@ def field_stat(args):
     if args.resultpath:
         relative_result_path = Path(args.resultpath)
     result_path = ens_path / relative_result_path
+    if not result_path.exists():
+        # Create the directory
+        logger.info(
+            "Create directory for results for field parameter statistics: "
+            f"{result_path}"
+        )
+        result_path.mkdir()
```
alifbe, line 368: We had discussion about this within subscript team regarding creating output path. I think majority of the code in subscript will trigger a warning that user should ensure the output path is already exist. There is discussion about removing the feature of automatic folder creating at some point.

https://github.com/equinor/subscript/blob/c6cef12c738dd3f80a7a2ca694c696c7081c9a03/src/subscript/prtvol2csv/prtvol2csv.py#L126

src/subscript/field_statistics/field_statistics.py
```diff
@@ -790,13 +851,13 @@ def write_mean_stdev_nactive(
             param_name=param_name,
         )
 
-    logger.info(f"Write parameter: {name_mean}")
+    logger.info(f"  Write parameter: {name_mean}")
```
alifbe, line 854: Any reason to have leading space here and a few more below?

src/subscript/field_statistics/field_statistics.py
```diff
@@ -967,50 +1007,120 @@ def get_specifications(input_dict, ertbox_size, ert_config_path):
...
 
-    key = "zone_conformity"
-    if key in input_dict:
-        zone_conformity = input_dict[key]
-    else:
-        raise KeyError(f"Missing keyword:  {key} specifying conformity per zone.")
-    check_zone_conformity(zone_code_names, zone_names_used, zone_conformity)
+    if not ertbox_size:
+        # ertbox size does not exist, read it from this scripts config file instead
+        print("ERTBOX size is not defined, need to get it from the config file")
```
alifbe, line 1013: I think it's best to user logger instead of printing it to screen. Especially for forward model or workflow job.

src/subscript/field_statistics/field_statistics.py
```diff
@@ -967,50 +1007,120 @@ def get_specifications(input_dict, ertbox_size, ert_config_path):
...
+    zone_code_names = None
     param_name_dict = None
-    key = "continuous_property_param_per_zone"
+    disc_param_name_dict = None
+    key = "geogrid_fields"
     if key in input_dict:
-        param_name_dict = input_dict[key]
-    check_param_name_dict(zone_code_names, param_name_dict)
+        use_geogrid_fields = True
+        geogrid_fields_dict = input_dict[key]
```
alifbe, line 1040: I would argue that the code will be more clear if you just use the key value directly i.e. `geogrid_fields` instead of putting it into `key` variable.

src/subscript/field_statistics/field_statistics.py
```diff
@@ -1291,17 +1409,19 @@ def calc_stats(
                             sum_total_active = np.ma.sum(sum_active) / nreal
                             sum_total_code = np.ma.sum(number_of_cells) / nreal
                             fraction = sum_total_code / sum_total_active
-                            txt1 = f"Average number of active cells: {sum_total_active}"
+                            txt1 = (
+                                f"  Average number of active cells: {sum_total_active}"
+                            )
```
alifbe, line 1414: I don't see txt1, txt2, txt3 and txt4 being used anywhere else. Would it be sufficient that you put it's content directly into logger.info?

src/subscript/field_statistics/field_statistics.py
oddvarlia: I can of course replace it with an Error message if it does not exist, but it is easy to forget making the directory.
When running as a HOOK_WORKFLOW in ERT, a FORWARD model to create the directory will be, required. When running it from the workflow menu in ERT, the user have to first go to the scratch and make the directory manually.

I can also add a warning message that the directory is created if it is automatically created, and keep the automatic creating of the directory.

src/subscript/field_statistics/field_statistics.py
oddvarlia: I can fix this

src/subscript/field_statistics/field_statistics.py
berland: It is considered better practice not to create directories implicitly, there are some downsides to it. Some code in subscript++ has had that feature previously, but it is typically deprecated or removed.

src/subscript/field_statistics/field_statistics.py
oddvarlia: I can fix this

src/subscript/field_statistics/field_statistics.py
oddvarlia: I can do that.

src/subscript/field_statistics/field_statistics.py
oddvarlia: The main reason why I usually assign a key string to a variable is that it is used multiple times. So if the key text string is to be changed for some reason, I can do it only in one line. In the case here it is referred two times only so not so important to assign it to a variable.

oddvarlia: I keep it as it is where it is referred to 3 times or more.

src/subscript/field_statistics/field_statistics.py
oddvarlia: OK, so I raise an Error then if not existing.

alifbe (approved):
LGTM

PR #763, opened 2024-12-05
Simplified FMU observation setup

anders-kiaer:
Follow-up of the good work done by Daniel in #704.


src/subscript/genertobs_unstable/__init__.py
alifbe, line 1: This is probably what caused an error in testing. `genertobs_unstable` should have its own `__init__.py`

anders-kiaer:
Scout will add native support directly in Ert in upcoming milestone.

PR #765, opened 2024-12-11
Use memoryconservative mode for WF CSV_MERGE

berland:
When run as a workflow job/ert plugin this has been observed through logging to fail due to lack of memory.

When run as a workflow job, it is probable that there is one CSV for every realization that is to be loaded, thus the memory conservative option probably makes sense to have on by default.

Keeping the faster option as default for usages as a forward model step.


berland:
https://fmu-dashboard-fdeqekb8chekcqhu.neu.grafana.azure.com/d/ae0bpm0169ou8e/traces?from=1731324844410&to=1733916844410&var-operation_id=dbd6018074a3166d77e6c6c04d5dea03&orgId=1&viewPanel=2

PR #771, opened 2025-02-11
Fix docs and tests after deprecation of runeclipse

berland:


eivindjahren (approved):

berland:
Verified manually to pass tests onprem.

PR #784, opened 2025-04-22
Replace sys.exit() with raise exception for workflow jobs

### LINKED ISSUE #783: csv_merge should not do system.exit(1)
csv_merge is in run both as a standalone tool and as a workflow in Ert. When running as a workflow inside Ert, it should not do `system.exit(1)`, but rather throw an exception to allow Ert to handle it (as of now, it will take down Ert abruptly)

https://github.com/equinor/subscript/blob/c7edbbfca4b95491cfbb40b234d824915dbb7aaf/src/subscript/csv_merge/csv_merge.py#L307

alifbe:
Close #783 

Also implement changes to other workflow jobs `csv_stack` and `field_statistics`

### DIFF
--- a/src/subscript/field_statistics/field_statistics.py
+++ b/src/subscript/field_statistics/field_statistics.py
@@ -15,6 +15,7 @@ Result:
 import argparse
 import copy
 import logging
+import sys
 from pathlib import Path
 
 import ert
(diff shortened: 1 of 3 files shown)


src/subscript/field_statistics/field_statistics.py
```diff
@@ -330,20 +329,20 @@ def main():
...
+        raise FileNotFoundError("No such file: " + args.configfile)
 
     config_file = args.configfile
     config_dict = read_field_stat_config(config_file)
     field_stat = config_dict["field_stat"]
 
     # Path to FMU project models ert/model directory (ordinary CONFIG PATH in ERT)
     if not Path(args.ertconfigpath).exists():
-        sys.exit("No such file:" + args.ertconfigpath)
+        raise FileNotFoundError("No such file: " + args.ertconfigpath)
```
rnyb, line 340: Change to f-string: `raise FileNotFoundError(f"No such file: {args.ertconfigpath}")`
And same for the other two

berland:
Looks fine to me. I have checked that the user experience running csv_merge on two empty csv files on the command line is not changed after this change.

Copilot comments are quite good, but probably not critical to do. The command line users sees the correct error message in the terminal anyways, and it should also be visible when running as a workflow in Ert even though the explanatory text is not in the exception itself.

berland (approved):

PR #788, opened 2025-04-22
Add missing Ert integration tests

### LINKED ISSUE #787: Add missing Ert integration tests
These forward model steps defined by subscript do not have an integration test to verify that the configuration
of the forward model step is valid when running Ert:

* casegen_upcars
* eclcompress
* ecldiff2roff
* eclgrid2roff
* eclinit2roff
* interp_relperm
* merge_unrst_files
* welltest_dpds

These should be in place in order to reliably solve #709

berland:
Resolves #787

### DIFF
--- a/tests/test_eclcompress.py
+++ b/tests/test_eclcompress.py
@@ -688,3 +688,43 @@ def test_file_is_binary(byte_sequence, expected, tmp_path):
     else:
         Path("foo.txt").write_bytes(byte_sequence)
         assert file_is_binary("foo.txt") == expected
+
+
+@pytest.mark.integration
+def test_ert_integration_with_defaulted_files(tmpdir):
+    pytest.importorskip("ert")
+    os.chdir(tmpdir)
+    grid_dir = Path("simulations/realization-0/iter-0/eclipse/include/grid")
+    grid_dir.mkdir(parents=True)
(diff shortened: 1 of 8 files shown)


tests/test_eclcompress.py
```diff
@@ -688,3 +688,43 @@ def test_file_is_binary(byte_sequence, expected, tmp_path):
...
+        f"""
+        NUM_REALIZATIONS 1
+        RUNPATH .
+        FORWARD_MODEL ECLCOMPRESS(<FILES>={tmpdir / "filelist"})
+    """,
+        encoding="utf-8",
+    )
+
+    subprocess.run(["ert", "test_run", "--disable-monitor", ert_config], check=True)
+    assert "PORO\n  2*1 /\n" in Path("poro.grdecl").read_text(encoding="utf-8")
```
rnyb, line 730: The numbers tested would be more related to fipnum rather than poro (i.e. name change). However, for testing purpose it doesn't matter at all so feel free to ignore comment.

rnyb (approved):
LGTM

tests/test_eclcompress.py
berland: Right! Since porosity valued at 1 is technically possible, I will keep it :)

PR #789, opened 2025-04-23
Use pytest.importorskip instead of ad-hoc code

berland:


alifbe (approved):
That's quite neat. 👍 

LGTM

Issue #800, opened 2025-07-03
Field statistics config file is missing specification of the name of the geogrid


oddvarlia:
The field_statistics script take as input a yaml config file. This config file is missing a keyword for specification of the name of the geogrid that is used as prefix for 3D parameters realization on scratch. Today the name 'geogrid' is hardcoded, but is not a part of FMU standard to always call it 'geogrid'.

PR #801, opened 2025-07-03
Added new keyword to define geomodel name

oddvarlia:
The FIELD_STATISTICS script did not specify name of the geogrid in its config file and used 'geogrid' as hardcoded name. This is now fixed and a new keyword 'geogrid_name' is introduced to make this script more flexible.

Closes: #800

### DIFF
--- a/src/subscript/field_statistics/field_statistics.py
+++ b/src/subscript/field_statistics/field_statistics.py
@@ -615,7 +615,7 @@ def read_ensemble_realization(
     geogrid_name,
 ):
     realization_path = Path(f"realization-{realization_number}/iter-{iter_number}")
-    grid_path = Path("share/results/grids/" + geogrid_name.strip() + ".roff")
+    grid_path = Path("share/results/grids/" + geogrid_name + ".roff")
     file_path_grid = Path(ensemble_path) / realization_path / grid_path
     if file_path_grid.exists():
         grid = xtgeo.grid_from_file(file_path_grid, fformat="roff")
@@ -624,7 +624,7 @@ def read_ensemble_realization(
         return None, None, None
 
     property_path = Path(
-        "share/results/grids/" + geogrid_name.strip() + f"--{property_param_name}.roff"
+        "share/results/grids/" + geogrid_name + f"--{property_param_name}.roff"
     )
     file_path_property = Path(ensemble_path) / realization_path / property_path
     property_param = xtgeo.gridproperty_from_file(file_path_property, fformat="roff")
@@ -645,7 +645,7 @@ def read_geogrid_realization(
     geogrid_name,
 ):
     realization_path = Path(f"realization-{realization_number}/iter-{iter_number}")
-    grid_path = Path("share/results/grids/" + geogrid_name.strip() + ".roff")
+    grid_path = Path("share/results/grids/" + geogrid_name + ".roff")
     file_path_grid = Path(ensemble_path) / realization_path / grid_path
     if file_path_grid.exists():
         grid = xtgeo.grid_from_file(file_path_grid, fformat="roff")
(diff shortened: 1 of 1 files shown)


src/subscript/field_statistics/field_statistics.py
```diff
@@ -608,17 +612,20 @@ def read_ensemble_realization(
     iter_number,
     property_param_name,
     zone_code_names,
+    geogrid_name,
 ):
     realization_path = Path(f"realization-{realization_number}/iter-{iter_number}")
-    grid_path = Path("share/results/grids/geogrid.roff")
+    grid_path = Path("share/results/grids/" + geogrid_name.strip() + ".roff")
```
alifbe, line 618: I think you don't need to strip `geogrid_name` anymore since you already do it when you read it from yaml file?

src/subscript/field_statistics/field_statistics.py
```diff
@@ -875,13 +890,14 @@ def ertbox_to_geogrid_statistics(
...
+    geogrid_name,
     param_name=None,
     facies_name=None,
 ):
     if param_name:
-        geogrid_stat_name = f"geogrid--{statistics_name}_{param_name}"
+        geogrid_stat_name = f"{geogrid_name.strip()}--{statistics_name}_{param_name}"
     if facies_name:
-        geogrid_stat_name = f"geogrid--{statistics_name}_{facies_name}"
+        geogrid_stat_name = f"{geogrid_name.strip()}--{statistics_name}_{facies_name}"
```
alifbe, line 900: Same here, no need to strip.

src/subscript/field_statistics/field_statistics.py
```diff
@@ -1042,6 +1064,13 @@ def get_specifications(input_dict, ertbox_size, ert_config_path):
         use_geogrid_fields = True
         geogrid_fields_dict = input_dict["geogrid_fields"]
 
+        key = "geogrid_name"
+        if key in geogrid_fields_dict:
+            geogrid_name = geogrid_fields_dict[key]
+            geogrid_name = geogrid_name.strip()
+        else:
+            raise KeyError(f"Missing keyword:  {key} specifying name of geomodel grid.")
+
```
alifbe, line 1073: I propose to have this optional to keep compatibility with current user setup. New user will have possibility to specify `geogrid_name`, but current user can still use the default value `geogrid`.

```suggestion
        geogrid_name = geogrid_fileds_dict.get("geogrid_name", "geogrid").strip()
```

alifbe (commented):
I leave a few comments and one suggestion. 

The main point is to ensure that user can still use existing setup without specifying `geogrid_name` in their configuration file but can customize it if they want to.

src/subscript/field_statistics/field_statistics.py
oddvarlia: fixed

src/subscript/field_statistics/field_statistics.py
```diff
@@ -608,17 +612,20 @@ def read_ensemble_realization(
...
     if file_path_grid.exists():
         grid = xtgeo.grid_from_file(file_path_grid, fformat="roff")
         subgrids = grid.subgrids if grid.subgrids else None
     else:
         return None, None, None
 
-    property_path = Path(f"share/results/grids/geogrid--{property_param_name}.roff")
+    property_path = Path(
+        "share/results/grids/" + geogrid_name.strip() + f"--{property_param_name}.roff"
+    )
```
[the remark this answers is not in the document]
oddvarlia: fixed

src/subscript/field_statistics/field_statistics.py
oddvarlia: fixed

src/subscript/field_statistics/field_statistics.py
oddvarlia: OK Fixed

PR #802, opened 2025-07-07
Support the OPM Flow keyword PYACTION

vkip:


alifbe (changes requested):
Cool 👍

The changes looks good to me. Could you please add a test case as well?

PR #811, opened 2025-09-16
`prtvol2csv` Handle Eclipse dumping out July as JLY

### LINKED ISSUE #810: Date parsing bug in prtvol2csv
Observed in logs:

```
Traceback (most recent call last):
  File "/prog/res/komodo/2025.09.01-py311-rhel8-numpy1/root/bin/prtvol2csv", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/prog/res/komodo/2025.09.01-py311-rhel8-numpy1/root/lib64/python3.11/site-packages/subscript/prtvol2csv/prtvol2csv.py", line 354, in main
    simvolumes_df = currently_in_place_from_prt(prt_file, args.fipname)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/prog/res/komodo/2025.09.01-py311-rhel8-numpy1/root/lib64/python3.11/site-packages/subscript/prtvol2csv/prtvol2csv.py", line 215, in currently_in_place_from_prt
    initial_date_object = datetime.strptime(
                          ^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/_strptime.py", line 567, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.11/_strptime.py", line 349, in _strptime
    raise ValueError("time data %r does not match format %r" %
ValueError: time data ' 5 JLY 1994' does not match format '%d %b %Y'
```

alifbe:
Closes #810 

Eclipse printing out JLY instead of JUL for July.

### DIFF
--- a/src/subscript/prtvol2csv/prtvol2csv.py
+++ b/src/subscript/prtvol2csv/prtvol2csv.py
@@ -211,6 +211,8 @@ def currently_in_place_from_prt(
             if date_matcher_ecl.search(line) is not None:
                 line_split = line.split("*")
                 initial_date_string = line_split[0].rstrip().split("  ")[-1]
+                # Handle Eclipse dumping out July as JLY instead of JUL
+                initial_date_string = initial_date_string.replace("JLY", "JUL")
                 initial_date_object = datetime.strptime(
                     initial_date_string, "%d %b %Y"
                 ).date()


berland:
Would like to see a test ensuring that `JLY` and the leading space will work in the future.

alifbe:
> Would like to see a test ensuring that `JLY` and the leading space will work in the future.

Good point. I have now added a minimum test for it.

src/subscript/prtvol2csv/prtvol2csv.py
```diff
@@ -205,6 +205,7 @@ def currently_in_place_from_prt(
     date_matcher_ecl = re.compile(r"^\s{2}REPORT\s{3}0")
     # Flow: Look for "Report step  0" at start of line
     date_matcher_flow = re.compile(r"^Report step\s{2}0")
+    initial_date_found = False
```
berland, line 208: Related to the test?

alifbe: Not really, I was trying to solve warning from pylance. 😅

tests/test_prtvol2csv.py
```diff
@@ -183,6 +183,23 @@ def test_prtvol2csv(tmp_path, mocker):
     pd.testing.assert_frame_equal(dframe, expected)
 
 
+def test_correct_parsing_date(tmp_path):
+    shutil.copy(TEST_PRT_DATADIR / "DROGON_FIPNUM.PRT", tmp_path / "DROGON_FIPNUM.PRT")
+    os.chdir(tmp_path)
+    # Replace "REPORT   0     1 JAN 2018" with "REPORT   0     5 JLY 2018"
+    prt_path = tmp_path / "DROGON_FIPNUM.PRT"
+    prt_text = prt_path.read_text(encoding="utf8")
+    prt_text = prt_text.replace(
```
berland, line 192: This is a slightly unsafe test, because if the `replace() `function fails to find your search string, `prt_text` will go through unmodified and no error/warning emitted. So if the underlying PRT file changes, this test will become defunct, but noone will notice. 

If you first assert that the string you are replacing is present, then you will avoid that problem.

src/subscript/prtvol2csv/prtvol2csv.py
berland: This is a kind 'may be Unbounded'-warning? Is there a ruff issue that that can be added to similarly require this? (it is a good find, but should probably go in its own commit with a corresponding git message)

alifbe: I can't see to find ruff rule for this. But I agree that it should be in its own commit.

PR #823, opened 2025-10-13
Remove unused pylint comments and cleanup

larsevj:
jinja2 is used directly in casegen_upcar
subscript-internal is now completor and has its own logging setup


alifbe (approved):
Great job 👍🎉

PR #833, opened 2025-10-30
Copy test function from pyscal instead of importing it

larsevj:
Seems unnecessary for pyscal to package and expose these test utilities.
Follow up (needed for) https://github.com/equinor/pyscal/pull/483


alifbe (approved):
LGTM 👍

PR #834, opened 2025-11-10
Fixup typing issues for mypy

larsevj:
Due to changes in resfo and pyscal


src/subscript/interp_relperm/interp_relperm.py
```diff
@@ -233,7 +242,7 @@ def make_interpolant(
     interp_param: dict[str, float],
     satnum: int,
     delta_s: float,
-) -> pyscal.WaterOilGas:
+) -> pyscal.WaterOilGas | pyscal.GasWater:
```
alifbe: Why is it needed?

PR #835, opened 2025-11-11
Add checking for rhel9 get_os_arch

andreas-el:
Currently the `rhel9` is set to use the same binaries as `rhel8`.

Verified on azure nodes running `rhel8` and `rhel9`.


larsevj (approved):
:+1:

PR #849, opened 2026-02-13
Disable ruff rule RUF069

larsevj:
See https://docs.astral.sh/ruff/rules/float-equality-comparison/

Might be worth enabling it later on, but seems to be quite prone to flagging "too much", cannot compare any input to 0.0 for instance.


alifbe (approved):
Agree, let's postpone it for now. 👌

PR #867, opened 2026-06-08
ci(workflows): add explicit permissions to GitHub Actions workflows

rnyb:
- codecov and linting workflows: Set `contents:read` 
- subscript workflow: Introduces a new deploy-docs job, uploads/downloads a docs artifact, renames the main job (subscript -> test), and changes the Git push strategy

### DIFF
--- a/.github/workflows/subscript.yml
+++ b/.github/workflows/subscript.yml
@@ -1,7 +1,7 @@
 name: subscript
 
 permissions:
-  contents: write
+  contents: read
 
 on:
   push:
@@ -24,7 +24,7 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
-  subscript:
+  test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/subscript.yml
```diff
@@ -83,21 +85,45 @@ jobs:
...
+      - name: Upload documentation artifact
+        if: matrix.python-version == '3.12' && success()
+        uses: actions/upload-artifact@v4
+        with:
+          name: docs-html
+          path: build/docs/html
+
+  deploy-docs:
+    needs: test
+    if: github.repository_owner == 'equinor' && github.ref == 'refs/heads/main'
```
alifbe, line 107: Do we need to check for `matrix.python-version` as well here?

rnyb: I don't think so, deploy-docs has no matrix — it runs as a single job. The Python version filtering is handled in the test job where only the python-version == '3.12' combination uploads the documentation artifact. The deploy-docs job simply downloads that one artifact and deploys it.

alifbe: I see, that makes sense 👍

alifbe (approved):
Looks good to me 👍

PR #872, opened 2026-06-12
Fixes for two CI issues

alifbe:
Fixes for two CI issues:
- Linting with new RUFF complains that we should not use `autouse=True` for the fixture. Remove it the flag fixes the issue since the tests actually explicitly specify the fixture.t
- Matplotlib 3.11.0 changes the plotting slightly, causing `summaryplot` tests to fail. Update the baseline to fix it.


rnyb (approved):
👍

PR #874, opened 2026-06-15
FIX: Restore `restartthinner` printout behavior without --quiet

### LINKED ISSUE #873: restartthinner is not displaying intended output
Due to https://github.com/equinor/subscript/commit/e949b0d17f3628b168d5a543fb4be1cd903d4db2 outputs from dry-runs are never printed to the terminal, as the default log-level is WARNING.

alifbe:
Closes #873

### DIFF
--- a/src/subscript/restartthinner/restartthinner.py
+++ b/src/subscript/restartthinner/restartthinner.py
@@ -2,7 +2,6 @@
 
 import argparse
 import datetime
-import logging
 import shutil
 import subprocess
 import tempfile
@@ -13,9 +12,7 @@ import numpy as np
 import pandas as pd
 from resdata.resfile import ResdataFile
 
-from subscript import __version__, getLogger
-
-logger = getLogger(__name__)
+from subscript import __version__
 
 DESCRIPTION = """
 Slice a subset of restart-dates from an E100 Restart file (UNRST)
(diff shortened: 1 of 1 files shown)


src/subscript/restartthinner/restartthinner.py
```diff
@@ -177,26 +177,25 @@ def restartthinner(
...
-        for idx, rstidx in enumerate(restart_indices):
-            slicepresent = "X" if rstidx in slicerstindices else ""
-            logger.info(
-                "%4d  %s  %s",
-                rstidx,
-                datetime.date.strftime(restart_dates[idx], "%Y-%m-%d"),
-                slicepresent,
-            )
-        logger.info("-----------------------")
+    logger.info("Selected restarts:")
```
berland: I am not sure there is any need for the user to see the log level and the logger name for each line. I would use `print` here too. Check the user experience in the terminal and choose what looks best.

alifbe: Fair enough. I have now change it to use `print` instead. 👍

src/subscript/restartthinner/restartthinner.py
```diff
@@ -252,6 +253,8 @@ def main() -> None:
         parser.error("Provide the UNRST file, not the DATA file")
     if args.quiet:
         logger.setLevel(logging.WARNING)
+    else:
+        logger.setLevel(logging.INFO)
```
berland: There is no logging statements left in the file now, so you may remove all lines related to `logger`

alifbe: True, I have now updated the code accordingly.

longudq-noquotes.sch
```diff
@@ -0,0 +1,5 @@
+UDQ
+
+DEFINE FU_PWRI1 WWIR A_01 i WWIR A_02 + WWIR A_03 + WWIR A_04 + WWIR A_05 + WWIR A_06 + WWIR A_07 + WWIR A_08 + WWIR A_09 + WWIR A_10  + WWIR A_11 + WWIR A_12 + WWIR A_13 + WWIR A_14 + WWIR A_15/
```
berland: Assuming this was not intended to be in the commit :)

alifbe: Yes, remnant from other test that is fixed in different PR.

berland (approved):
👍🏻

PR #878, opened 2026-07-23
Fix issue with ruff 0.15.22

alifbe:
Add support for ruff 0.15.22

- Use descriptive ruff ignore rules instead of its code number
- Ignore ruff preview feature noqa-comments (at least until it's out of preview)


rnyb (approved):
👍

PR #883, opened 2026-08-28
Escape backslashes in RST doc strings

berland:
The unescaped backslashes were interpreted by RST while rendering, which was not the intention. Also remove a poorly rendered caption that reads better without it.

### DIFF
--- a/src/subscript/hook_implementations/forward_model_steps.py
+++ b/src/subscript/hook_implementations/forward_model_steps.py
@@ -59,8 +59,8 @@ It has capability to:
 
   DEFINE <CASEGEN_CONFIG_FILE>      <RUNPATH>/model.yaml
   DEFINE <CASEGEN_ECLIPSE_TEMPLATE> <CONFIG_PATH>/../input/config/eclipse.tmpl
-  FORWARD_MODEL CASEGEN_UPCARS(<CONFIG>=<CASEGEN_CONFIG_FILE>, \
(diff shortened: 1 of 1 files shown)


src/subscript/hook_implementations/forward_model_steps.py
```diff
@@ -867,21 +866,21 @@ def documentation() -> ForwardModelStepDocumentation | None:
...
-                 <OUTPUTDIRECTORY>=dst, <OUTFILESSUFFIX>=OP_1_1,
+   FORWARD_MODEL WELLTEST_DPDS(<ECLBASE>, <WELLNAME>=OP_1, <PHASE>=GAS, \\
+                 <BUILDUP_NR>=2, <OUTPUTDIRECTORY>=dst, <OUTFILESSUFFIX>=OP_1_1, \\
                  <GENOBS_RESULTFILE>=OP_1_dpdt_bu2_saphir.txt )
 
 Then set-up of GEN_DATA can be
 ::
 
-   GEN_DATA DPDT_SIM INPUT_FORMAT:ASCII REPORT_STEPS:1
+   GEN_DATA DPDT_SIM INPUT_FORMAT:ASCII REPORT_STEPS:1 \\
```
eilskra, line 883: Is it needed here as well?

berland: Yes, here it was missing altogheter and the syntax was plain wrong, as RESULT_FILE is an option to GEN_DATA

eilskra (approved):
🚀

PR #885, opened 2026-09-15
fix(fmuobs): use Series.mask for NA string handling in date conversion

rnyb:
Newer pandas-stubs reject Series.replace() when replacing string values
with np.nan (a float), causing mypy/basedpyright to fail on py313 CI.

- Replace .replace([...], np.nan) with .mask(isin(...), np.nan)
- Introduce local na_strings list for the missing-value markers
- Clarify docstring on replaced missing value strings ("NaT", "NaN", "nan")

Added minor extension to test.


berland:
There is a test `test_fmuobs_writers.py::test_convert_dframe_to_str()` that perhaps can be amended to also catch this.

rnyb:
> There is a test `test_fmuobs_writers.py::test_convert_dframe_to_str()` that perhaps can be amended to also catch this.

@berland I added a small extension to the test, does that cover your comment? https://github.com/equinor/subscript/pull/885/changes/643be97ed8c523b3d49f2d5aa86f0906857fc6c0

berland (approved):
👍🏻

